PR #3, opened 2026-07-02
feat(flow-nodes): add get operation for reading node config

cu4nt0m:
## What

Adds a `get` operation to `manage_flow_nodes` so the agent can read a single node's full config — including a **code** node's TypeScript. `list` returns only `id`/`type`/`label`/`parentId`, so before this there was no way to read a node's code back. This enables a **read-then-edit** workflow.

Note: create/update/delete for code nodes already existed — the gap was the *read* side. An agent couldn't inspect or safely edit code it didn't write. Now it can.

## Changes

- **`get` operation** → `GET /chart/nodes/{id}`, shaped by a new `filterFlowNodeDetail` that keeps `config` but strips the server-computed `transpiled` output (can be ~200k chars) while keeping `hasError`.
- **Transpile-error surfacing.** Code nodes are TypeScript, transpiled server-side at save. `hasError` = the code failed to compile.
  - `get` surfaces it via `_hints.warning`.
  - `update` surfaces it too — but the PATCH response echoes the input config *without* the post-transpile `hasError`, so after a code edit the handler does a read-back `GET` to detect it reliably.
- **Docs corrected.** The code node was mislabeled "JavaScript" — it's TypeScript. Updated the flow-nodes skill + registry summary, and documented the runtime objects (`input`/`context`/`actions`/`profile`) and the read-then-edit workflow.

## Testing

- Unit tests: `get` (returns config, strips `transpiled`, missing-nodeId error, `hasError` hint) and `update` (warns on transpile failure, silent on clean save). 332 pass.
- **Verified live** against a real Cognigy tenant via the plugin's MCP tools: full lifecycle on a throwaway code node (create → get → break → get → fix → get → delete), confirmed in the UI at each step. The update-path warning was found dead during smoke testing (PATCH echo lacks `hasError`) and fixed with the read-back.


tmccaughey:
Reviewed — clean, well-tested. No correctness bugs. Minor nits (non-blocking):

- `filters.ts` `filterFlowNodeDetail`: `raw?.config` is guarded but `raw.type`/`rid(raw)` aren't — throws if GET returns an empty body.
- `mock` asymmetry: `update` strips `mock.hasError`, `get` keeps it. Harmless, just inconsistent.
- Code-node update now does 3 sequential API calls (GET → PATCH → read-back GET). Fine — gated on code edits only.

LGTM to merge.

tmccaughey (approved):

PR #4, opened 2026-07-03
feat(setup): cognigy-setup CLI for GUI credential config

cu4nt0m:
## Problem

Installing the plugin from the **GUI** (desktop/web app) completes cleanly but does **not** show the `userConfig` credential prompt that the terminal install does. The MCP server then launches without `COGNIGY_API_BASE_URL` / `COGNIGY_API_KEY` and every tool call fails. Claude Code exposes no `claude plugin config set` CLI and the keychain naming for `sensitive` fields is undocumented, so there is no supported non-interactive way to populate the credentials.

## Solution

A `cognigy-setup` bin (run via `npx @cognigy/plugin-engine cognigy-setup`) that collects the two values and writes them to `~/.cognigy-plugin/config.json`:

- **API key entry is masked** (raw-mode, asterisks) — never echoed to the terminal or a shell history.
- File is written **owner-only** (dir `0700`, file `0600`), re-`chmod`ed on overwrite.
- `loadConfig()` reads the file **only when the matching env var is unset**, per-field — so the terminal keychain flow is untouched and env vars keep precedence.
- Non-interactive/scripted use: `--api-base-url` / `--api-key` flags.

The secret lives only in an owner-only file under `$HOME` — not in shared `~/.claude/settings.json` and not in any chat transcript.

## Changes

- `src/setup.ts` — interactive wizard + flags
- `src/userConfigFile.ts` — read/write helpers (tolerant read → `{}` on missing/malformed)
- `src/config.ts` — on-disk fallback + setup-guided error messages
- `package.json` — `bin: { cognigy-setup }`
- `README` — GUI install fallback section + Configuration note

## Testing

- `npx tsc --noEmit` clean; full suite green.
- New tests: `userConfigFile.test.ts` (write/read, `0600` perms incl. overwrite, malformed/array handling) and `config.test.ts` (env precedence, file fallback, per-field override, setup-guided errors).
- Packed the tarball, installed it into a scratch project, confirmed `node_modules/.bin/cognigy-setup` resolves and writes the file at `chmod 600`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,16 @@ The plugin is supported by **Claude Code** and **Codex** today; more clients wil
 
 On enable, Claude Code prompts for your **Cognigy API base URL** (default `https://api-trial.cognigy.ai`) and **API key** (Cognigy.AI → User Menu → My Profile → API Keys). The key is stored in your system keychain. Requires [Node.js 20+](https://nodejs.org).
 
+#### If you install from the GUI and aren't prompted for credentials
+
+The desktop/web app installs the plugin cleanly but does not always show the credential prompt. If the Cognigy tools report a missing API key, run the setup command once in any terminal:
+
+```
+npx @cognigy/plugin-engine cognigy-setup
+```
+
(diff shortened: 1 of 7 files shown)


tmccaughey (approved):
LGTM

PR #10, opened 2026-07-20
fix(plugin): launch engine via npx so Desktop marketplace sync works

cu4nt0m:
## Problem

Adding the `Cognigy/cognigy-plugin` marketplace in the **standalone Claude Desktop app** fails: *"Marketplace sync failed. Check the repository URL and try again."* / *"Failed to add marketplace."* It has always worked in Claude Code.

## Root cause

Claude Desktop syncs marketplaces **server-side** and security-scans plugin files. The committed launcher `plugin/bin/launch.mjs` runs `npm install` + spawns the engine at boot (fetch-and-execute remote code) — the scanner rejects it and fails the **entire** marketplace add. Claude Code doesn't scan, so `claude plugin validate --strict` and marketplace add always passed there.

Bisected empirically in throwaway repos (single variable): a skills+agents-only plugin syncs on Desktop; adding back `launch.mjs` reproduces the failure. Ruled out: xApps content, an external-script template, the `mcpServers` block, subdir `source: "./plugin"`, marketplace name, `owner.url`, proxy/network, Desktop version, cache.

## Fix

Drop the launcher file; run the engine directly from the manifest:

```
npx -y -p @cognigy/plugin-engine@<version> cognigy-mcp
```

A manifest command is not a scanned file → Desktop sync passes and **skills load**. On Desktop the `platform` connector stays credential-less (a no-op) as before; MCP tools still come from the installer-wired direct connector. On Claude Code the engine runs via npx (terminal PATH resolves it), version-pinned for lockstep.

## Changes

- **`package.json`** — re-add the `cognigy-mcp` bin (`dist/index.js`).
- **`plugin/.claude-plugin/plugin.json`** — `mcpServers.platform` → npx command; **delete `plugin/bin/launch.mjs`**.
- **`-p` form everywhere** — two bins mean npx can't pick a default, so a bare `npx <pkg> cognigy-setup` throws *"could not determine executable to run"* (the #7 regression). Updated README, `src/config.ts` error messages, `desktopLauncher.ts` hint, `src/setup.ts`, and the one-bin test → asserts both bins.
- **`scripts/sync-plugin-version.mjs`** — also rewrites the engine pin inside the npx args, keeping plugin ↔ engine versions in lockstep.
- Docs: `CLAUDE.md`, `docs/TESTING.md`.

## Tradeoffs

Reintroduces an npx PATH dependency (fine on Claude Code's terminal PATH; the Desktop connector was already a no-op) and drops the launcher's explicit first-run-race guard + Windows `.cmd` handling (npx manages its own).

## Verification

- Typecheck clean; `npm test` 378 passing.
- Local smoke test: `npx -p <packed-tgz> cognigy-mcp` selects the right bin (two bins present) and responds to an MCP `initialize` with **clean stdout** (npm/npx noise → stderr; logger → stderr) — stdio stream uncorrupted.
- Version-sync verified: rewrites both `version` and the engine pin.
- Desktop marketplace add of an equivalent config (npx, no bin file) confirmed to sync.

## Release note

Ships in one release: semantic-release publishes the new engine version **with** the `cognigy-mcp` bin, and `sync-plugin-version.mjs` pins the plugin's npx command to that same version — closing the loop.


PhotonicDev (approved):

tmccaughey (approved):
lgtm

PR #11, opened 2026-07-20
feat(installer): node-check bootstrap, desktop-only path, lifecycle subcommands

cu4nt0m:
Overhauls the install UX. Design doc: `docs/plans/install-ux.md`.

## What & why

Building on #10 (which made the Desktop marketplace add work), this closes the gaps in how people actually install:

### 1. Node-check bootstrap (repo-root `install.sh` / `install.ps1`)
New headline install:
```
bash <(curl -fsSL https://raw.githubusercontent.com/Cognigy/cognigy-plugin/main/install.sh)
```
- Uses `bash <(curl …)` **process substitution**, not `curl | sh`, so stdin stays attached and the installer's masked API-key prompt still works.
- **Detects Node 20+ only** — if missing, prints nodejs.org / nvm guidance and exits. **Never auto-installs** (per decision), so the machine's toolchain is untouched.
- Then hands off to the existing `npx … cognigy-setup`.
- **Placed at repo root, not under `plugin/`** — a curl+exec script in the marketplace-scanned plugin tree could re-trip the #10 Desktop scan.

### 2. Desktop-only users (no `claude` CLI)
Post-Apr-2026, Desktop has a **Code tab** (embedded Claude Code) that shares `~/.claude` with the CLI, but a Desktop-only user has **no `claude` on PATH**, and `/plugin` is CLI-only. The old no-CLI fallback (print `/plugin` commands) was a dead end for them. Now the message covers both: terminal Claude Code (`/plugin`) **and** the Desktop Code-tab **GUI plugin browser** (`+` → Plugins → Add plugin → `Cognigy/cognigy-plugin`) for skills+agents.

### 3. Lifecycle subcommands
`cognigy-setup status | update | uninstall` (default `install`):
- `status` — installed surfaces vs latest npm; flags drift.
- `update` — `claude plugin update` (or prints it); notes Desktop auto-updates on boot.
- `uninstall` — reverses both surfaces; `--purge` also clears `~/.cognigy-plugin`; confirm prompt.

### 4. Auto-update propagation
After a CLI install, print the one-time marketplace auto-update toggle. **Verified there's no non-interactive CLI flag** for it (`marketplace add` has only `--scope`/`--sparse`), so we guide rather than hand-edit Claude's internal JSON.

### 5. Menu relabel
Client labels reflect the redesign (Claude Code = CLI + Desktop Code tab; Claude Desktop = chat connector).

## Migration (no forced migration; nobody breaks)
- **Desktop connector users** auto-update the engine on next restart (unchanged mechanism).
- **Claude Code plugin users** stay on their pinned version (cached `launch.mjs` still works) until they enable auto-update / `/plugin update`. Item 4 fixes propagation going forward.

## Verification
- Typecheck clean; `npm test` **382 passing** (+4: update/uninstall/marketplace-remove arg builders, `removeDesktopServerEntry`).
- `install.sh` syntax-checked; `cognigy-setup status` and unknown-command smoke-tested locally.

## Open items (flagged in the doc)
1. Verify a real Desktop marketplace add still succeeds with root `install.sh` present (a root file *might* also be scanned).
2. Confirm the current Desktop Code-tab plugin-browser UI steps (screenshots drift).
3. Bootstrap hosting: raw `main` vs a NiCE vanity domain.


PhotonicDev (approved):
LGTM!

PR #13, opened 2026-07-23
feat(flow): render operation — visualize flows (ASCII + mermaid + HTML)

cu4nt0m:
## What

Closes the visualization gap when working on flows over MCP. Adds `manage_flow_nodes { operation: "render" }` — the engine **deterministically** serializes the flow chart into multiple views in plain code. The model never re-derives the graph, so no hallucinated edges and ~0 token cost.

## Views, by client capability

| View | Field | Renders in |
|---|---|---|
| ASCII tree | `ascii` | every client incl. terminal |
| Mermaid graph | `mermaid` | Desktop / web / **mobile** — as a native Mermaid artifact |
| Rich HTML page | `htmlUrl` (with `writeHtml:true`) | browser (tmp file, no publishing) |
| Legend | `legend` | shape/edge key for exactly the elements present |

- **ASCII** walks `relations[].next` as a vertical chain and nests `children` as tree branches.
- **Mermaid**: solid arrow = `next`, dotted `contains` = `children`; shapes per node type; traversal-order emission nudges Default leftmost / End last.
- **HTML**: self-contained page written to `os.tmpdir()`; **mermaid is inlined so it renders fully offline** (no CDN). Opens in the browser automatically. **No artifact publishing.**
- `focus=<nodeId>` (or an array) highlights one or more nodes (e.g. ones just created/edited) with readable text → partial-render / diff-style feedback.

## How the mermaid is delivered (important)

The mermaid string must be delivered as a **native Mermaid/diagram artifact** — the "Download and open · MERMAID" card. That is the only form that renders natively and zoomably, including on phones. Hard rules enforced via `_hints`:
- Do **not** wrap it in an HTML page / widget / generic "visualize" connector (breaks mobile).
- Do **not** paste it as an inline ```mermaid fenced block (plain text / tiny thumbnail).
- Keep commentary, `legend`, and `ascii` outside the artifact.

## Legend — minimal, dynamic

- Lists **only** the shapes/edges present in the specific flow.
- In-diagram (mermaid): a compact left→right strip; the shape **is** the key (a diamond labeled "Branch"), never a shape name; quiet transparent styling.
- HTML panel: draws the actual shapes as inline SVG, theme-aware.

## Render suggestion after mutations

After a flow mutation (create/update/delete node), the result carries a minimal one-time `_hints.renderSuggestion` so the model briefly offers to render the updated flow — never unprompted, never repeated.

## Design notes

- Serializers in `src/render/flowRender.ts` are **pure** (JSON in / string out) — fully unit-tested.
- **`next` is a single id string in the live API** (not an array); `nextIds()` normalizes both. A real captured chart fixture guards this regression.
- Handler merges the chart endpoint (topology + `preview`) with the node-list endpoint (`label`).
- `withHints` tells the model how to present each field per client.

## Testing

- `src/__tests__/flowRender.test.ts` — chain walk, nesting, multi-focus, label/preview/type fallback, empty flow, cycle guard, deterministic output, in-diagram legend, drawn-SVG HTML legend, offline-inline HTML, real string-`next` API fixture.
- Manual demo: `npx tsx scripts/render-demo.ts`.
- Full suite green (411 passed, 3 skipped); `tsc --noEmit` clean; prettier clean.

## Installer polish (same branch)

- Claude Code install ends with a clear `✅ — all set` (tools + skills + agents; no further steps).
- Claude Desktop chat steps simplified: `Customize → Plugins → Add → Add marketplace` (dropped stale "Personal plugins" wording).

## Review follow-ups

- **Slimmed the `manage_flow_nodes` tool description** (~4050 → ~1850 chars). It now carries only the tool contract (operations, params, the one mermaid-delivery hard rule) plus a pointer to the skill. The workflow detail moved to where the architecture wants it: the `flow-nodes` skill gained a **render** section (ascii/mermaid/legend/focus/writeHtml + delivery rules) and auto-loads on render/visualize intent; `instructions.ts` carries a terse cross-tool render rule for skill-less clients (Codex).
- **Windows `start`**: pass an empty title arg (`start "" <file>`) so paths containing spaces open reliably.
- **CDN HTML path**: switched from `startOnLoad:true` to `startOnLoad:false` + explicit `await mermaid.run()`, matching the offline path — no async-import render race.


src/tools/definitions.ts
```diff
@@ -770,7 +770,7 @@ After creating, use talk_to_agent to test.`,
...
...eep commentary/legend/ascii outside it). Pass focus=<nodeId> (or an array of node IDs) to highlight one or more nodes (e.g. ones you just created/edited) so the user sees how their change fits. Pass writeHtml=true to also write a self-contained rich HTML graph to a tmp file on the local machine (this server runs on the same computer as the user) and get back `htmlUrl`/`htmlPath` — the file is already on that machine, so just hand the user the link to open in a browser; do NOT try to fetch, upload, or regenerate it. It opens in the browser automatically (openInBrowser defaults to true; pass false to only get the link back). Read-only.\n\nSUPPORTED NODE TYPES: the authoritative set is defined by the server node registry and returned in the error if you pass an unsupported type. Common types: say, question, ifThenElse, lookup, setSessionContext, code, goTo, sleep, httpRequest. xApp nodes: initAppSession, showXAppAdaptiveCard, showXAppHtml, setXAppState, getXAppSessionPin (initAppSession must precede other xApp nodes).\n\nFor AI Agent tool nodes (knowledge, send_email, mcp, http), use create_tool / update_tool instead — they handle the required parent-child node wiring automatically.',
```
PhotonicDev: Umm, this is getting really long, do you think that some part of this should be a skill rather than part of the tool description?

cu4nt0m: @PhotonicDev Yes pretty valid imo too. Will trim it down and source it out to skills and instructions, as we still are expected to give a good enough experience in skill-less platforms, so having a brief sum in instructions would be nice.

cu4nt0m: Good call — done. Trimmed the `manage_flow_nodes` description to just the tool contract (operations + params + the one mermaid-delivery hard rule) plus a pointer to the `flow-nodes` skill (~4050→1850 chars). Moved the workflow detail (placement, branching, and a new render section) into `plugin/skills/flow-nodes/SKILL.md`, and put the cross-tool render rule in `instructions.ts` for skill-less clients (Codex). Also fixed the two Copilot nits in the same push. 5d0b8xx.

🤖 Generated but reviewed by me!

PhotonicDev (approved):
LGTM!

PR #21, opened 2026-08-07
feat(tools): rename flows/projects/agents with DELETE_ prefix instead of deleting

dshire:
## Summary

- **Nothing is hard-deleted anymore** by `delete_resource` for the protected types: flows, projects, and agents are renamed to `DELETE_<name>` to mark them for manual deletion in the Cognigy UI (idempotent: already-marked resources are left unchanged and reported as `alreadyMarked`)
- **Endpoints are deactivated, not deleted**: agent deletion (default `cascade: true`) and flow deletion PATCH every endpoint referencing the flow to `active: false` — taking the agent/flow offline reversibly (the "Enable Endpoint" toggle in the UI) — then rename flow + agent; `cascade: false` renames only the agent
- Add `project` as a new `delete_resource` resourceType with rename-only semantics; the warning states explicitly that the rename takes nothing offline (flows, endpoints, and agents inside remain live)
- Warnings are composed from actual per-step outcomes with `cascade.failed` as the source of truth — the response never claims the agent is offline unless endpoint deactivation fully succeeded
- `create_ai_agent`'s failure rollback keeps hard-deleting the resources it created in the same failed call; endpoint/llm_model/knowledge_store/function/tool deletion is unchanged
- Behavior note: since agents are never deleted, re-running `create_ai_agent` with the original name after a soft-delete creates a new agent alongside the `DELETE_`-prefixed one

## Changes

| --- | --- |
| `src/tools/handlers.ts` | Replace `cascadeDeleteAgent` with `softDeleteAgent` (deactivate endpoints, rename flow + agent); add `deactivateEndpointsForFlow` / `renameForDeletion` / `markForDeletion` helpers; route flow/project/agent through rename paths; outcome-composed warnings |
| `src/tools/definitions.ts` | Rewrite `delete_resource` description for the new semantics; add `project` to the enum; update `cascade` description |
| `src/schemas/tools.ts` | Add `project` to `deleteResourceSchema` enum |
| `src/instructions.ts` | Document rename-instead-of-delete + endpoint deactivation in the always-on baseline instructions |
| `plugin/skills/troubleshooting/SKILL.md` | Replace stale cascade-delete note with the new rename/deactivation semantics |
| `src/__tests__/tools.test.ts`, `src/__tests__/schemas.test.ts` | Coverage: flow/project rename, endpoint deactivation on both paths, already-marked idempotency (result shape + no phantom `renamed` entries), agent soft cascade, `cascade: false`, resource-identifying errors, 200-char truncation, schema accepts `project` |
| `README.md` | Update `delete_resource` row in the tools table |

## Test plan

Manual — smoke-tested against a real trial tenant (both previously-unused API calls): `PATCH /v2.0/projects/{id} { name }` and `PATCH /v2.0/endpoints/{id} { active: false }` behave as expected; flow soft-delete deactivated its endpoint (visible as the Enable Endpoint toggle off in the UI), agent soft-delete on the already-marked flow reported `alreadyMarked` without a phantom rename, project rename warned that contents stay live.


src/tools/handlers.ts
```diff
@@ -2466,13 +2545,24 @@ export class ToolHandlers {
...
+        id: agentId,
+        cascade: {
+          deleted,
+          renamed,
+          failed: failed.length > 0 ? failed : undefined,
+        },
+      },
+      {
+        warning:
+          "Agents and flows cannot be deleted via this plugin. The agent and its flow were renamed with the DELETE_ prefix to mark them for manual deletion; endpoints were permanently deleted to take the agent offline.",
```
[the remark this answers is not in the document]
dshire: Fixed in 0a272aa. The warning is now composed from actual outcomes: it only claims endpoints were deleted when endpoint enumeration ran and nothing failed, and otherwise states that cleanup was incomplete / the flow could not be resolved and the agent may still be reachable, pointing at `cascade.failed`.

src/tools/handlers.ts
```diff
@@ -2370,32 +2393,88 @@ export class ToolHandlers {
...
+    url: string,
+    currentName: string | undefined,
+  ): Promise<{ name: string; alreadyMarked: boolean }> {
+    const name = currentName ?? "";
+    if (name.startsWith(DELETE_PREFIX)) {
+      return { name, alreadyMarked: true };
+    }
+    const newName = `${DELETE_PREFIX}${name}`;
+    await this.apiClient.patch(url, { name: newName });
+    return { name: newName, alreadyMarked: false };
```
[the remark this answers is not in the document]
dshire: Fixed in 0a272aa. `renameForDeletion` now throws on an empty/blank name instead of patching `DELETE_`, and truncates the prefixed name to the 200-char limit the schemas enforce.

src/tools/handlers.ts
```diff
@@ -2466,13 +2545,24 @@ export class ToolHandlers {
...
-    return {
-      deleted: allSucceeded,
-      resourceType: "agent",
-      id: agentId,
-      cascade: { deleted, failed: failed.length > 0 ? failed : undefined },
-    };
+      failed.length === 0 && renamed.includes(`agent:${agentId}`);
+    return withHints(
+      {
+        markedForDeletion: allSucceeded,
```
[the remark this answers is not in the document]
dshire: Fixed in 0a272aa. `markedForDeletion` now reflects whether the agent rename succeeded (independent of endpoint/flow outcomes), and the warning is built from the actual per-step results with `cascade.failed` as the source of truth.

src/tools/handlers.ts
```diff
@@ -2370,33 +2396,99 @@ export class ToolHandlers {
...
+    );
+    await this.apiClient.patch(url, { name: newName });
+    return { name: newName, alreadyMarked: false };
+  }
+
+  /**
+   * Flows and projects cannot be deleted via the plugin — rename them to
+   * DELETE_<name> so a human can delete them manually in the Cognigy UI.
+   */
+  private async markForDeletion(
```
cu4nt0m, line 2687: **Medium:** A soft-deleted flow (or project) stays fully live. Unlike the agent path, this renames only — endpoints referencing the flow keep serving traffic under the new `DELETE_` name, and nothing inside a renamed project is taken offline. The warning never mentions this, so an LLM (or user) will believe the resource is out of service.

Suggestion: at minimum, extend the warning with "the flow may still be reachable via its endpoints" (and the project equivalent); ideally enumerate/delete referencing endpoints the way `softDeleteAgent` does.

src/tools/handlers.ts
```diff
@@ -2370,33 +2396,99 @@ export class ToolHandlers {
...
+   * 3. Rename the flow to DELETE_<name>
+   * 4. Rename the agent to DELETE_<name>
    */
-  private async cascadeDeleteAgent(agentId: string): Promise<any> {
+  private async softDeleteAgent(agentId: string): Promise<any> {
     const deleted: string[] = [];
+    const renamed: string[] = [];
     const failed: { resource: string; error: string }[] = [];
 
     const resolved = await resolveFlowForAgent(this.apiClient, agentId);
```
cu4nt0m, line 2786: **Minor:** Partial-failure retry can lose the flow. If the flow rename succeeds but the agent rename fails, a retry of `delete_resource { resourceType: "agent" }` goes through `resolveFlowForAgent`, whose strategy 3 matches `${agent.name} Flow` — the flow is now `DELETE_My Agent Flow`, so it won't match when the agent has no direct `flowId`. The retry then silently skips endpoint cleanup (the `endpointsChecked: false` warning does fire, so it's at least visible).

src/tools/handlers.ts
```diff
@@ -2454,25 +2546,70 @@ export class ToolHandlers {
       }
     }
 
-    // Step 3: delete the agent
+    // Step 3: rename the agent (agents are never deleted)
     try {
-      await this.apiClient.delete(`/v2.0/aiagents/${agentId}`);
-      deleted.push(`agent:${agentId}`);
+      await this.renameForDeletion(`/v2.0/aiagents/${agentId}`, agent?.name);
+      renamed.push(`agent:${agentId}`);
```
cu4nt0m: **Minor:** On an idempotent re-run, `renameForDeletion` returns `alreadyMarked: true` without issuing a PATCH, but this still pushes to `renamed[]` — so `cascade.renamed` claims a rename that didn't happen. Also, the agent path's result omits the `name` and `alreadyMarked` fields that the flow/project path returns; aligning the shapes would help clients.

cu4nt0m (commented):
Reviewed the full diff. The outcome-composed warnings, idempotency handling, and the 200-char cap look solid, and the create_ai_agent rollback correctly still hard-deletes via direct API calls. Two non-code items:

- The PR currently conflicts with main and needs a rebase before merge.
- `PATCH /v2.0/projects/{id} { name }` is only exercised by mocks — the project rename endpoint was never used before this PR. Please run one manual smoke test against a real tenant before merging (per the repo rule that mocks may not match real API shapes).

Inline comments below for the code findings (1 medium, 4 minor).

src/tools/handlers.ts
```diff
@@ -2370,33 +2396,99 @@ export class ToolHandlers {
...
+   * Idempotent: an already-prefixed name is left unchanged.
+   */
+  private async renameForDeletion(
+    url: string,
+    currentName: string | undefined,
+  ): Promise<{ name: string; alreadyMarked: boolean }> {
+    const name = currentName ?? "";
+    if (!name.trim()) {
+      throw new Error(
+        "Cannot mark resource for deletion: it has no name to prefix.",
```
cu4nt0m: **Minor:** This error is uncaught on the `cascade: false` and flow/project paths and carries no resource type or id — "Cannot mark resource for deletion: it has no name to prefix" leaves the caller guessing which resource failed. Consider passing type/id into `renameForDeletion` (or wrapping at the call sites) so the message identifies the resource.

src/tools/handlers.ts
```diff
@@ -2370,33 +2396,99 @@ export class ToolHandlers {
...
+      );
+    }
+    if (name.startsWith(DELETE_PREFIX)) {
+      return { name, alreadyMarked: true };
+    }
+    const newName = `${DELETE_PREFIX}${name}`.slice(
```
cu4nt0m, line 2604: **Nit/edge:** Two names longer than 193 chars that differ only past that point truncate to the same `DELETE_` name. If Cognigy enforces unique names, the second rename fails (it is recorded in `cascade.failed`, so it degrades gracefully — just noting the edge).

src/tools/handlers.ts
dshire: Fixed in e6aa5d3 (refined in d50688b) — and we went further than the suggested warning: flow soft-delete now runs the same endpoint step as the agent path, via a shared deactivateEndpointsForFlow helper. Since nothing is hard-deleted anymore, the old reason for deleting endpoints (the API refused to delete a referenced flow) is gone, so both paths now PATCH endpoints to active: false instead — reversible via the Enable Endpoint toggle — and the flow result gains a cascade block. Projects stay rename-only, but the warning now states explicitly that the rename takes nothing offline and contents remain live. Smoke-tested against a real trial tenant: the endpoint shows up deactivated (not deleted) in the UI.

src/tools/handlers.ts
dshire: Fixed in e6aa5d3 (wording refined in d50688b): renameForDeletion's alreadyMarked result is now honored — idempotent re-runs no longer push a phantom entry into cascade.renamed, and the warning says 'already carried the DELETE_ prefix' instead of claiming a rename. The agent result also returns name and alreadyMarked, aligning its shape with the flow/project paths.

src/tools/handlers.ts
dshire: Fixed in e6aa5d3: renameForDeletion now takes a resource label, so the error reads e.g. 'Cannot mark flow 6a8e… for deletion: it has no name to prefix' on every path (cascade: false and flow/project included).

src/tools/handlers.ts
dshire: Acknowledged, leaving as a known edge for now: the retry stays visible via the endpointsChecked-false warning ('endpoints could not be resolved — the agent may still be reachable'), and agents created via create_ai_agent carry a direct flowId, so the name-based strategy-3 fallback only matters for hand-wired agents. With endpoints now merely deactivated rather than deleted (see the main thread), the worst case of a skipped endpoint step is also milder than before.

src/tools/handlers.ts
dshire: Acknowledged — real edge, but it degrades into cascade.failed as you note, and >193-char names differing only past that point should be vanishingly rare. Leaving as-is.

dshire:
Addressed the review — summary of what changed since:

- **Semantic change**: endpoints are now **deactivated** (`PATCH { active: false }`, reversible via the Enable Endpoint toggle) instead of permanently deleted, on both the agent and the new flow path. Nothing in this tool hard-deletes flows/projects/agents *or* their endpoints anymore; the PR description is updated accordingly.
- **Rebased onto main** (resolving against the new `manage_snapshots` instructions); CI is green.
- **Manual smoke test done** against a real trial tenant, covering both previously mock-only API calls: `PATCH /v2.0/projects/{id} { name }` renamed the project as expected, and `PATCH /v2.0/endpoints/{id} { active: false }` deactivated the endpoint (visible in the UI, endpoint preserved). Also verified the idempotent agent re-run (`alreadyMarked`, no phantom `cascade.renamed` entry) and the project warning stating contents stay live.

Inline replies on the individual findings: the medium and two minors are fixed (e6aa5d3, d50688b); the retry-loses-flow edge and the truncation-collision nit are acknowledged and left as-is.

cu4nt0m (approved):

PhotonicDev (approved):
LGTM

PR #22, opened 2026-08-07
feat(tools): support AWS Bedrock LLM provider in setup_llm

dshire:
## Summary

- Add `awsBedrock` provider support to `setup_llm` — Amazon Nova, Claude-on-Bedrock, and Titan embedding models can now be configured (follow-up promised in #20)
- Add AWS-specific credential handling: `accessKeyId` + `secretAccessKey` auto-creates an `AwsBedrockProvider` connection, `roleArn` an `AwsBedrockProviderIamRole` connection; `apiKey` is rejected for this provider with an actionable error
- Send the required `awsBedrock: { region, customModel? }` provider metadata and expose it in `llm_model` responses

**Stacked on #20** (`feat/setup-llm-openai-compatible`) — retarget to `main` after #20 merges.

## Changes

| File / Component                       | Description                                                                                                                                                  |
| -------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| `src/schemas/tools.ts`                 | `awsBedrock` in the provider enum; new `region`/`accessKeyId`/`secretAccessKey`/`roleArn` fields; provider-specific field rules generalized into a `SETUP_LLM_FIELD_PROVIDERS` map; Bedrock rules (region required, key-pair completeness, roleArn XOR keys, customModel↔custom-model coupling) |
| `src/tools/handlers.ts`                | Credential detection per provider; connection auto-creation picks type + fields (`AwsBedrockProvider` / `AwsBedrockProviderIamRole`); `awsBedrock` metadata in the create payload |
| `src/tools/definitions.ts`             | Provider enum, new parameters, and an AWS BEDROCK usage section in the `setup_llm` description                                                                 |
| `src/tools/filters.ts`                 | `llm_model` filter returns `awsBedrock` metadata                                                                                                               |
| `plugin/skills/llm-providers/SKILL.md` | Provider table row + AWS Bedrock section (supported model ids, credential options, example)                                                                    |
| `src/__tests__/*`                      | 11 new schema/handler tests (payload shape, both connection types, validation errors)                                                                          |

Connection types (`AwsBedrockProvider`, `AwsBedrockProviderIamRole`), metadata shape (`region`, `customModel?`), and the supported Bedrock model-id list were verified against the Cognigy.AI source. Not yet live-tested — needs an AWS account with Bedrock access.

## Follow-up

`googleGenAI` support lands in the next stacked PR.

## Test plan


src/tools/definitions.ts
```diff
@@ -153,14 +153,15 @@ export const tools: ToolDefinition[] = [
...
         },
         modelType: {
           type: "string",
           description:
-            "Model type string. Chat examples: 'gpt-4o', 'gpt-4o-mini', 'claude-sonnet-4-0', 'mistral-small-2503'. Embedding examples: 'text-embedding-3-small', 'text-embedding-3-large', 'text-embedding-ada-002', 'gemini-embedding-001'. For provider 'openAICompatible' this MUST be 'custom-model' (chat) or 'custom-embedding-model' (embedding); the real model name goes in customModel.",
+            "Model type string. Chat examples: 'gpt-4o', 'gpt-4o-mini', 'claude-sonnet-4-0', 'mistral-small-2503'. Embedding examples: 'text-embedding-3-small', 'text-embedding-3-large', 'text-embedding-ada-002', 'gemini-embedding-001'. For provider 'openAICompatible' this MUST be 'custom-model' (chat) or 'custom-embedding-model' (embedding); the real model name goes in customModel. For 'awsBedrock' use a Bedrock model id (e.g. 'amazon.nova-pro-v1:0') or 'custom-model' + customModel.",
         },
         name: {
           type: "string",
```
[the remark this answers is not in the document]
dshire: Fixed in 696b2de — the `name` description now matches the runtime fallback (`name || customModel || modelType`), noting customModel also applies to awsBedrock with modelType 'custom-model'.

src/tools/handlers.ts
```diff
@@ -1989,23 +2000,37 @@ export class ToolHandlers {
       }
```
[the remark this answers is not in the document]
dshire: Fixed in 696b2de — the hint is now provider-aware: for awsBedrock it suggests accessKeyId + secretAccessKey (or roleArn) instead of apiKey.

plugin/skills/llm-providers/SKILL.md
```diff
@@ -63,11 +64,46 @@ Notes:
...
 - When inspecting existing models via `list_resources` / `get_resource`, openAI-compatible models show `modelType: "custom-model"`; the real model name and endpoint are in the `openAICompatible` object (`customModel`, `baseCustomUrl`, `customAuthHeader`).
 
+## AWS Bedrock (awsBedrock)
+
+Use provider `awsBedrock` for models hosted on AWS Bedrock (Amazon Nova, Anthropic Claude on Bedrock, Titan embeddings).
+
+Required parameters:
+
+- `region`: AWS region of the Bedrock deployment, e.g. `us-east-1`
+- `modelType`: a Bedrock model id from Cognigy's supported list — chat: `amazon.nova-pro-v1:0`, `amazon.nova-lite-v1:0`, `amazon.nova-micro-v1:0`, `amazon.nova-2-lite-v1:0`, `anthropic.claude-3-5-sonnet-20240620-v1:0`; embedding: `amazon.titan-embed-text-v2:0`. For any other Bedrock model use `custom-model` and put the model id — or an inference profile id like `eu.anthropic.claude-sonnet-4-6` — in `customModel`.
```
cu4nt0m, line 74: `anthropic.claude-3-5-sonnet-20240620-v1:0` is no longer a valid choice. Cognigy's deprecations-and-removals table lists it as deprecated on 2026-03-01 and removed in Cognigy 2026.10.0, and AWS has end-of-lifed the model id as well.

Following this list on a current tenant means `setup_llm { provider: 'awsBedrock', modelType: 'anthropic.claude-3-5-sonnet-20240620-v1:0', ... }` is either rejected on POST or fails the connection probe, the model is auto-deleted with a credentials-oriented hint, and the auto-created AWS connection is left behind.

Suggest dropping this id (keep the four Nova ids and the Titan embedding model) and relying on the `custom-model` + inference-profile guidance already in this paragraph for Claude, e.g. `eu.anthropic.claude-sonnet-4-6`.

src/tools/handlers.ts
```diff
@@ -1989,23 +2003,37 @@ export class ToolHandlers {
...
     }
 
-    // If apiKey is provided, auto-create a Connection first
-    if (data.apiKey && !connectionRefId) {
+    // If inline credentials are provided, auto-create a Connection first
+    if (hasInlineCredentials && !connectionRefId) {
```
cu4nt0m, line 2007: Two related issues here:

1. `if (hasInlineCredentials && !connectionRefId)` silently drops `accessKeyId` / `secretAccessKey` / `roleArn` whenever `connectionId` is also passed, so the `AwsBedrockProviderIamRole` branch never runs in that case. The new docs list the three auth options as unordered alternatives with no precedence, so a caller asked to "switch this to IAM-role auth" will plausibly pass the existing key-pair `connectionId` plus `roleArn`. The call succeeds on the old key-pair connection with no warning, and the user believes IAM-role auth is in effect. Suggest rejecting (or at least warning on) inline credentials combined with `connectionId`.

2. The `connectionId` lookup above (around lines 1964-1969) never checks `match.type` against the provider, so an `OpenAIProvider` connection is accepted for `provider: 'awsBedrock'` and only fails at the probe, or never with `dangerouslySkipConnectionTest`. Now that one provider has two valid connection types, `PROVIDER_CONNECTION_TYPE`'s single string cannot express the right check. Suggest `Record<Provider, string[]>` and validating `match.type` for all providers.

There is currently no test covering `awsBedrock` + `connectionId`.

src/tools/handlers.ts
```diff
@@ -1989,23 +2003,37 @@ export class ToolHandlers {
...
+          fields: connectionFields,
         });
         connectionRefId =
           connection.referenceId || connection._id || connection.id;
       } catch (connError: any) {
         return withHints(
           { error: `Failed to create connection: ${connError.message}` },
           {
-            action: "Check API key and provider, then retry.",
+            action: "Check credentials and provider, then retry.",
```
cu4nt0m, line 2036: When the connection type is `AwsBedrockProviderIamRole` and the installation does not have the IAM-role feature flag (the default, per the SKILL text in this PR), `POST /v2.0/connections` fails deterministically with "This type is not enabled for your installation". The generic hint here ("Check credentials and provider, then retry.") contradicts the platform message passed through in `connError.message` and never suggests the documented fallback, so the model will re-ask the user to verify the ARN and retry the identical call.

Suggest branching the hint on `connectionType === 'AwsBedrockProviderIamRole'` and pointing at `accessKeyId` + `secretAccessKey` as the fallback.

src/schemas/tools.ts
```diff
@@ -96,30 +135,71 @@ export const setupLlmSchema = z
...
+          path: ["apiKey"],
+          message:
+            "Provider 'awsBedrock' does not use apiKey. Provide accessKeyId + secretAccessKey (access-key auth) or roleArn (IAM-role auth) instead.",
+        });
       }
-      if (
-        data.apiType !== undefined &&
-        data.provider !== "openAI" &&
-        data.provider !== "azureOpenAI"
+      if (data.roleArn && (data.accessKeyId || data.secretAccessKey)) {
```
cu4nt0m, line 157: Minor: this `if … else if` couples two independent validations, unlike every other check in the `awsBedrock` block. With `{ roleArn, accessKeyId }` (no `secretAccessKey`) the caller only sees "Provide either accessKeyId + secretAccessKey OR roleArn, not both"; after dropping `roleArn` it is told on a second round-trip that `secretAccessKey` is missing. Two plain `if`s would report both at once. If the coupling is intentional (to avoid steering toward completing the pair), a short comment would help.

src/tools/handlers.ts
```diff
@@ -1930,9 +1931,19 @@ export class ToolHandlers {
   async handleSetupLlm(args: any): Promise<any> {
     const data = schemas.setupLlmSchema.parse(args);
 
-    if (!data.apiKey && !data.connectionId) {
+    const hasInlineCredentials =
+      data.provider === "awsBedrock"
```
cu4nt0m, line 1935: Structural note rather than a bug: `handleSetupLlm` now special-cases `data.provider === "awsBedrock"` in five places (here, the error text at 1943, the action hint at 1981, the mutable `let connectionType` / `connectionFields` override at 2011-2013 with the literal `'AwsBedrockProviderIamRole'` inline while `PROVIDER_CONNECTION_TYPE` at 503 holds `'AwsBedrockProvider'`, and the `providerMeta` nested ternary at 2057), plus a bespoke `if (data.apiKey)` check in the schema because `apiKey` is the one field left out of `SETUP_LLM_FIELD_PROVIDERS`. The `?? data.provider` fallback at 2009 is unreachable against the closed zod enum.

The PR description names `googleGenAI` as the next stacked provider. With this shape that adds a sixth arm at each site plus another schema block, and missing one silently falls back to `apiKey` semantics (`connectionFields = { apiKey: data.apiKey! }` with `apiKey` undefined). A single per-provider descriptor table (connection types per auth mode, credential field sets, meta fields, credential hint) consumed by both handler and schema would make a new provider one table entry and put both Bedrock connection-type strings in one place.

src/tools/definitions.ts
```diff
@@ -132,7 +132,7 @@ export const tools: ToolDefinition[] = [
...
... the model is deleted and an error is returned — this prevents broken model references from silently breaking downstream flows.\n\nIf the provider's test endpoint is unreachable or returns a non-testable status, the model is kept but a warning is returned so you know connectivity could not be fully verified.\n\nIf dangerouslySkipConnectionTest is true, the connection test is not run at all; the model is kept and the response includes a warning that no connectivity check was performed. Only use this flag when you explicitly accept the risk that the created LLM might not be callable. Never use it to bypass a missing or cross-project connection.\n\nIf isDefault is true (the default), agents in the project will automatically use this LLM. If isDefault is false, you must explicitly assign it to the agent via update_ai_agent { aiAgentId, jobConfig: { llmProviderReferenceId: '<referenceId from this response>' } }.\n\nThe response includes the LLM's referenceId — use this value for jobConfig.llmProviderReferenceId if assigning manually.\n\nTo list existing LLMs: use list_resources { resourceType: 'llm_model', projectId }.\nTo delete: use delete_resource { resourceType: 'llm_model', id }.",
```
cu4nt0m, line 135: The supported Bedrock model-id list now lives in three copies that already disagree: `SKILL.md:74` (six ids), this description (three ids), and the schema error messages (`schemas/tools.ts:179` / `:187`). This file is also internally inconsistent: the AWS BEDROCK paragraph gives `eu.anthropic.claude-sonnet-4-6` as the `customModel` example while the `customModel` property description gives `anthropic.claude-sonnet-4-20250514-v1:0`.

Cognigy's Bedrock list changes with platform releases (`nova-premier` deprecated in 2026.13, `claude-3-5-sonnet` removed in 2026.10), so clients that load skills (Claude Code, Antigravity) and clients that only see this description (Codex) will be told different valid ids. Suggest keeping the list only in the `llm-providers` skill, having the description say "a Bedrock model id from Cognigy's supported list (see the llm-providers skill) or `custom-model` + `customModel`", and aligning the two `customModel` examples. The paragraph also adds roughly 200 tokens to every turn and largely restates the six new per-property descriptions.

src/tools/definitions.ts
```diff
@@ -153,19 +153,20 @@ export const tools: ToolDefinition[] = [
             "google",
```
cu4nt0m, line 153: The 7-value provider enum and the new `location` enum (`['region', 'geo', 'global']`) are hand-duplicated verbatim between the zod schema (`schemas/tools.ts:55-63`, `:74`) and this JSON `inputSchema` (here and at 212), and the provider list appears a third time inside the description string at 158. `toolParameters.test.ts` only checks structural shape, not enum parity, so a value added to one list and not the other is not caught in CI.

The repo already has the pattern to avoid this: `AUDIT_ACTORS` / `AUDIT_EVENT_TYPES` are exported `as const` from `schemas/tools.ts` and spread as `enum: [...AUDIT_ACTORS]` at `definitions.ts:379` / `:389`. Suggest exporting `SETUP_LLM_PROVIDERS` / `BEDROCK_LOCATIONS` the same way and using them in both `z.enum(...)` and the `inputSchema`.

cu4nt0m (commented):
## Review: AWS Bedrock support in `setup_llm`

Typecheck is clean, the touched Jest suites pass (357 tests), Prettier is clean, and there is no hand version bump. I checked the findings below against the Cognigy OpenAPI reference, the Bedrock provider docs page, the release notes and the deprecations table.

**Blocking (confirmed)**

1. `plugin/skills/llm-providers/SKILL.md:74` lists `anthropic.claude-3-5-sonnet-20240620-v1:0` as a supported Bedrock chat model. Cognigy's deprecations table records it as deprecated 2026-03-01 and removed in 2026.10.0, and AWS has end-of-lifed the id. On any current tenant the create call is rejected or the probe fails, the model is auto-deleted, and the auto-created AWS connection is left behind. See inline comment.
2. `src/tools/handlers.ts:2169` — the connection-test-failure hint still reads `Verify your API key and model type are correct, then retry.` This is the only credential wording in `handleSetupLlm` that was not made provider-aware (the siblings at 1943, 1981 and 2036 were), while the schema now hard-rejects `apiKey` for `awsBedrock`. A wrong AWS secret produces a dead-end loop: the hint asks for an API key, the retry with `apiKey` fails schema validation with "Provider 'awsBedrock' does not use apiKey". (Outside the diff, so no inline comment.)
3. `src/instructions.ts:29` still says "If an LLM needs to be created and no API key was provided, ASK for it". The same stale wording remains in the `apiKey` parameter description (`definitions.ts:174`), `plugin/agents/cognigy-agent-builder.md:16` ("ask the user for provider/model/API key"), `troubleshooting/SKILL.md:48` and `llm-providers/SKILL.md:120`. In a skill-less client (Codex) or inside the agent-builder subagent this steers the model to ask for an API key for Bedrock, which the schema then rejects — exactly the round-trip the updated tool description tries to prevent. Suggest generalising to "no credentials were provided" and adding an `awsBedrock` caveat to the `apiKey` description. (File not in the PR, so no inline comment.)
4. `src/tools/handlers.ts:2007` — inline credentials are silently discarded when `connectionId` is also passed, and the `connectionId` lookup never checks `match.type` against the provider. See inline comment.
5. `src/tools/handlers.ts:2036` — a `roleArn` connection-create failure on an installation without the IAM-role feature flag gets the generic "Check credentials and provider, then retry." hint. See inline comment.

**Non-blocking / hardening**

6. `src/tools/handlers.ts:2076` — the create payload sends `awsBedrock: { region, location?, geo?, customModel? }`. That object is not in the public OpenAPI spec for `POST /v2.0/largelanguagemodels` (no `awsBedrock` object, no Bedrock ids in the `modelType` enum), the PR description says it is not live-tested, and the later-added `location` / `geo` keys are not covered by the "verified against Cognigy source" note. The three new handler tests only assert that the handler echoes its own payload into the mocked `api.post`, so CI cannot catch a shape mismatch. Please verify against a live 2026.11+ tenant before merging.
7. `src/schemas/tools.ts:157` — two independent validations coupled with `else`; see inline comment.
8. `src/tools/handlers.ts:1935` — `awsBedrock` is special-cased in five places; see inline comment about a per-provider descriptor table before `googleGenAI` lands.
9. `src/tools/definitions.ts:135` — the Bedrock model-id list now exists in three inconsistent copies; see inline comment.
10. `src/tools/definitions.ts:153` — provider and `location` enums hand-duplicated between zod and the JSON schema; see inline comment.

Refuted candidates, for the record: rejecting `custom-embedding-model` for Bedrock matches the platform (Bedrock custom models are chat-only), and hard-rejecting `apiKey` alongside `connectionId` is a deliberate, tested design.

plugin/skills/llm-providers/SKILL.md
dshire: Fixed in ccaf3a7 — dropped `anthropic.claude-3-5-sonnet-20240620-v1:0`; Claude on Bedrock now goes through `custom-model` + an inference profile id, with a note that the list changes per release.

src/tools/handlers.ts
dshire: Fixed in ccaf3a7 — (1) the schema now rejects inline credentials combined with `connectionId` for every provider; (2) `PROVIDER_CONNECTION_TYPES` is a `Record<string, readonly string[]>` with both Bedrock types, and the matched connection's `type` is validated against it. Tests added for both Bedrock + `connectionId` paths.

src/tools/handlers.ts
dshire: Fixed in ccaf3a7 — the hint branches on `AwsBedrockProviderIamRole` and points at `accessKeyId` + `secretAccessKey` as the fallback.

src/schemas/tools.ts
dshire: Fixed in ccaf3a7 — two plain `if`s; a test asserts both issues are reported at once.

src/tools/handlers.ts
dshire: Deferring the descriptor-table refactor to the `googleGenAI` follow-up PR, where the third provider shape will drive the abstraction. In ccaf3a7 both Bedrock connection-type strings now live in one place (`PROVIDER_CONNECTION_TYPES`) and the unreachable `?? data.provider` fallback is gone.

src/tools/definitions.ts
dshire: Fixed in ccaf3a7 — the model-id list lives only in the llm-providers skill; the description and `modelType` property defer to it. Both `customModel` examples now use `eu.anthropic.claude-sonnet-4-6`, and the AWS BEDROCK paragraph is trimmed.

src/tools/definitions.ts
dshire: Fixed in ccaf3a7 — `SETUP_LLM_PROVIDERS` / `BEDROCK_LOCATIONS` are exported `as const` from `schemas/tools.ts` and spread into the `inputSchema`, same as `AUDIT_ACTORS`.

dshire:
Addressed the review in ccaf3a7. The out-of-diff items:

- **Connection-test-failure hint** (`handlers.ts`): now provider-aware — for `awsBedrock` it asks to verify `accessKeyId` + `secretAccessKey` (or `roleArn`), region, and model id.
- **"Ask for the API key" wording**: generalised to credentials in `src/instructions.ts`, `plugin/agents/cognigy-agent-builder.md`, `troubleshooting/SKILL.md` and `llm-providers/SKILL.md`; the `apiKey` parameter description now states Bedrock does not use it.

Inline threads have individual replies. Deferred: the per-provider descriptor table (to the `googleGenAI` PR) and live-tenant verification of the `awsBedrock` payload (still needs an AWS account with Bedrock access).

PhotonicDev (approved):
LGTM!

PR #26, opened 2026-08-13
fix: make credentials work on hosts without userConfig support (VS Code, Cursor)

dshire:
## Why

Found while testing VS Code / Copilot for the [plugin compatibility matrix](https://cognigy.atlassian.net/wiki/spaces/Engineering/pages/2684289044/Plugin+compatibility).

VS Code 1.125 installs our plugin properly: the marketplace works, all 13 skills load, both agents load, and it even reads the `mcpServers` block **inside** `plugin.json` (its docs only describe a separate `.mcp.json`, so this was a pleasant surprise). The server process starts. Then every tool call fails:

```
{"error":"Invalid URL","code":"ERR_INVALID_URL"}
```

`userConfig` is a **Claude Code extension** to the manifest format. Claude Code prompts for the values and substitutes them into `mcpServers.*.env`. Hosts implementing only the portable subset — VS Code / Copilot, Cursor — copy that text through verbatim, so the engine receives the literal string `"${user_config.cognigy_api_key}"`.

Those placeholders are **non-empty**, which broke two things in `loadConfig()`:

1. `COGNIGY_API_BASE_URL` reached axios as a base URL of `${user_config.cognigy_api_base_url}` → `ERR_INVALID_URL`.
2. Worse: they satisfied the "are the env vars set?" check and **shadowed the on-disk fallback**, so credentials written by `cognigy-setup` were ignored even when they existed.

There is no way to fix this from the host side. VS Code's own `${input:...}` secret prompt applies only to servers a user writes in their own `mcp.json`; there is no mechanism to inject a value into a server a *plugin* declares. So the fix has to be ours.

## What changed

**1. `src/config.ts` — treat unexpanded placeholders as unset**

A value that is entirely an unexpanded `${...}` is now ignored, so it falls through to `~/.cognigy-plugin/config.json`. The match is anchored to the whole trimmed value — a real key or URL is never wholly wrapped in `${…}`, so a genuine credential can't be discarded. The "not set" errors now also say when a placeholder was the cause, which otherwise reads as nonsense to someone staring at a manifest that plainly sets the value.

**2. `src/setup.ts` — new `other-hosts` install target**

Fix 1 alone wasn't enough: nothing wrote the fallback file on the happy path. `installClaudeCode` only wrote it when the `claude` CLI was **missing** or the install **failed** — so anyone who installed successfully into Claude Code (i.e. most people) had no fallback at all.

I deliberately did **not** make it always write. That would put a plaintext API key on disk for every user, including Claude-only users who today get keychain storage and nothing on disk — a silent security downgrade nobody asked for. Instead there's a third target in the client picker, and it is **never auto-detected**, so it is never pre-checked. Opt-in only.

It also prints the `npx: command not found` guidance (see below).

## Result

Install the plugin in VS Code → run `npx cognigy-setup` and tick **Other hosts** → tools, skills, and agents all work. No VS-Code-specific code, and the same file fixes Cursor and every other host with this limitation.

## How to verify

**Automated:**

```bash
npm test -- --runInBand      # 431 pass, 0 fail
npx tsc -p tsconfig.json --noEmit
npm run lint                 # 17 errors before and after — all pre-existing
                             # parse errors (test files are in tsconfig.test.json)
```

**The actual bug, before and after.** The unit tests mock `readUserConfigFile`, so they never prove the real file is read. This does:

```bash
FAKE=$(mktemp -d); mkdir -p "$FAKE/.cognigy-plugin"
cat > "$FAKE/.cognigy-plugin/config.json" <<'JSON'
{ "COGNIGY_API_BASE_URL": "https://api-trial.cognigy.ai", "COGNIGY_API_KEY": "from-the-file" }
JSON
cat > "$FAKE/probe.ts" <<'TS'
import { loadConfig } from "./src/config.js";
console.log("apiBaseUrl:", loadConfig().apiBaseUrl);
TS

HOME="$FAKE" \
  COGNIGY_API_BASE_URL='${user_config.cognigy_api_base_url}' \
  COGNIGY_API_KEY='${user_config.cognigy_api_key}' \
  npx tsx "$FAKE/probe.ts"
```

- On `main`: `apiBaseUrl: ${user_config.cognigy_api_base_url}` ← the bug
- On this branch: `apiBaseUrl: https://api-trial.cognigy.ai`

**The installer** (temp `HOME` so it can't touch your real config):

```bash
FAKE=$(mktemp -d)
HOME="$FAKE" npx tsx src/setup.ts --client other-hosts \
  --api-base-url https://api-trial.cognigy.ai --api-key TEST-KEY --yes
ls -la "$FAKE/.cognigy-plugin/"   # expect: dir 0700, config.json 0600
```

**End to end in VS Code:**

1. `"chat.plugins.enabled": true` and `"chat.plugins.marketplaces": ["Cognigy/cognigy-plugin"]` in settings.json
2. Extensions view → `@agentPlugins` → install **cognigy**
3. `npx cognigy-setup`, tick **Other hosts**
4. Restart VS Code, start the `platform` server, ask it to list your projects

If the server fails to start with `npx: command not found`, see the unresolved item below.

## Unresolved: `npx` resolution, cause not isolated

The manifest uses `"command": "npx"`, resolved from PATH. During testing on macOS with node installed via nvm, we saw it **both** fail and succeed, and did not isolate why:

- ❌ A `cognigy-manual` server in VS Code's **user `mcp.json`** failed with `npx: command not found`. Switching to an absolute path (`~/.nvm/versions/node/v24.15.0/bin/npx`) fixed it.
- ✅ The **plugin's own `platform` server**, with `"command": "npx"` untouched, started fine and discovered all 16 tools.

Two candidate explanations, untested:

1. **Launch method** — a Dock/Spotlight-launched app doesn't source your shell profile, so nvm's bin dir is invisible; a terminal launch inherits it. The two observations may simply have been launched differently.
2. **Config source** — VS Code may resolve the shell environment for plugin-declared servers but not for ones in the user `mcp.json` (or the reverse).

I originally wrote this up as a general "GUI hosts can't see nvm" product gap affecting every host in the compatibility matrix. **That was overstated** — one failure isn't enough to support it, and the plugin server working contradicts it. Recording it here as an open question rather than a known bug.

Worth isolating separately, since it decides whether anything needs fixing at all. `nvm` is common enough that if the failure mode is real and launch-dependent, it's worth knowing; if VS Code resolves shell env reliably, there's nothing to do. This PR only adds conditional troubleshooting text to the installer output, shown when a user actually hits the error.

## Notes for review

- Precedence is unchanged: real env vars still win, the file is still only a fallback. Claude Code keeps using the keychain and never reads the file.
- The fallback file is `0600` in a `0700` dir (unchanged; `writeUserConfigFile` re-tightens both in case they pre-existed).
- Two commits, `fix:` + `feat:`, so semantic-release picks a minor bump. Per convention I have not touched either version field.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,16 @@ The plugin ships its own connector (`platform`), but on **Claude Desktop chat** 
 
 > 🔄 **Claude Desktop users:** the connector keeps its engine current on every restart, but see [Staying up to date](#staying-up-to-date) for how updates work and how to check versions.
 
+### Step 3 — Other hosts (VS Code, Cursor, …): install the plugin yourself
+
+Pick **Other hosts** in the installer (or pass `--client other-hosts`) and it writes your credentials to `~/.cognigy-plugin/config.json` (`chmod 600`). Nothing else is wired — these hosts install the plugin themselves:
+
+**VS Code / Copilot** — in `settings.json`, set `"chat.plugins.enabled": true` and add `"chat.plugins.marketplaces": ["Cognigy/cognigy-plugin"]`, then open the Extensions view, search `@agentPlugins`, and install **cognigy**. Restart afterwards. You get tools, skills, and agents.
+
+**Leave the plugin's credential fields alone.** Only Claude Code implements `userConfig`, the manifest extension that prompts for an API key and substitutes it into the server's environment. Every other host copies the manifest text through verbatim, so the engine would receive the literal `${user_config.cognigy_api_key}`. It treats such placeholders as unset and reads the file above instead — which is why the installer writes it.
+
+> **`npx: command not found` when the server starts?** Your Node is probably from nvm/fnm/volta, whose bin directory a GUI-launched host may not have on `PATH`. Quit the host completely and relaunch it from a terminal, or point the MCP entry at an absolute path (`$(which npx)`).
+
 <details>
 <summary>Scripting / CI, and manual install</summary>
 
@@ -113,6 +123,8 @@ npx -y -p @cognigy/plugin-engine@latest cognigy-setup \
   --client claude-code --client claude-desktop \
   --api-base-url https://api-trial.cognigy.ai --api-key <key>
 ```
+
+`--client` is repeatable; valid values are `claude-code`, `claude-desktop`, and `other-hosts`. Omit it in interactive mode to pick from the menu.
 
 **Manual install (Claude Code)** — instead of the installer:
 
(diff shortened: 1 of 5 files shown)


cu4nt0m (approved):

PhotonicDev (approved):
LGMT

PR #27, opened 2026-08-14
feat(installer): native Antigravity plugin support (IDE + agy CLI)

PhotonicDev:
## Summary

- Adds `--client antigravity` to `cognigy-setup`, giving **Google Antigravity** (IDE + `agy` CLI + SDK) the full surface — 16 tools, 13 skills, 2 agents — from one command.
- Ships it as a **native Antigravity plugin** registered via `agy plugin install`, rather than scattering files into global config, so `agy plugin uninstall cognigy-plugin` cleanly reverses it and the user's own MCP servers are never touched.
- Fixes a latent packaging gap: `plugin/skills` and `plugin/agents` never shipped in the npm package, so any non-marketplace installer had nothing to copy.

Everything below was verified against a **live Antigravity install**, not the docs — the published docs are stale on several points (see Notes).

### How it installs

The installer stages a plugin under `~/.cognigy-plugin/antigravity-plugin` and hands it to `agy plugin install`, which copies it to `~/.gemini/config/plugins/cognigy-plugin/` and records it in `import_manifest.json`. `~/.gemini/config` is shared by the IDE, the CLI and the SDK, so one install serves all three.

```
~/.gemini/config/plugins/cognigy-plugin/
├── plugin.json          name + version + description
├── mcp_config.json      cognigy (stdio engine) + cognigy-docs (serverUrl)
├── skills/<id>/SKILL.md 13 workflow skills
└── agents/<id>/agent.md 2 subagents
```

When `agy` is not on PATH, the installer copies the plugin into place itself and sets `plugins.cognigy-plugin.enabled` in `config.json` — the same shape Antigravity's bundled plugins use.

### Credentials

The API key is **not** written into any `mcp_config.json`. That file is shared, hand-edited and routinely pasted into bug reports — and in practice it is created world-readable (`644`). Credentials go to `~/.cognigy-plugin/config.json` (`0600`), which `loadConfig()` already reads whenever the env vars are absent, so the plugin's server entry carries no `env` at all. Any stale global `cognigy` entry left by an earlier or hand-rolled setup is removed so only one engine boots.

This converges with the approach in #18 and #26 — see Related work.

## Changes

| File / Component                            | Description                                                                                                                         |
| ------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------- |
| `src/install/antigravity.ts`                | **New.** Stages the plugin, registers it via `agy` (with a direct-write fallback), plus detect/status/update/uninstall helpers      |
| `src/setup.ts`                              | New `antigravity` client wired into the menu, `--client`, detection, and the `install` / `status` / `update` / `uninstall` commands |
| `scripts/copy-assets.mjs`                   | Copies `plugin/skills` + `plugin/agents` into `dist/plugin-assets/` so they ship in the npm package                                 |
| `src/install/desktopLauncher.ts`            | Generalised comments + one user-facing string; the launcher is now shared with Antigravity (no behaviour change)                    |
| `src/__tests__/installAntigravity.test.ts`  | **New.** 14 tests over manifest/MCP shape, staging layout, the `config.json` flag and legacy-entry cleanup                          |
| `src/__tests__/install.test.ts`             | `parseFlags` accepts `antigravity`                                                                                                  |
| `README.md`                                 | Antigravity install path + a details block on exactly what gets written                                                             |
| `.claude/CLAUDE.md`, `docs/ARCHITECTURE.md` | Architecture notes, including the doc-vs-reality gotchas below                                                                      |

## Test plan

Automated:

- `npm test` — 436 passed, 0 failed (3 pre-existing skips)
- `npm run check:manifest`
- `npx prettier --check` on all changed files
- `npx eslint src/install src/setup.ts` — clean

Manual (against a real Antigravity install on macOS):

- `agy plugin validate <staged dir>` → `13 skills · 2 agents · 2 mcpServers` all processed
- Ran the installer end to end → registered via the `agy` path; `agy plugin list` shows the plugin with `skills, agents, mcpServers`
- Booted the plugin's MCP server exactly as Antigravity launches it (command + args from the installed `mcp_config.json`) **with no env vars**: 16 tools listed and instructions loaded, confirming credentials resolve from the `0600` file
- Confirmed the global `~/.gemini/config/mcp_config.json` is left as `{"mcpServers": {}}` — the plugin's config is read in place

Not covered: Windows and Linux paths for `agy` detection are code-reviewed only; the fallback (no-`agy`) branch is unit-tested but has not been exercised against a real Antigravity that lacks the CLI.

## Notes — docs vs. reality

Worth recording, since the published Antigravity docs disagree with a current install:

- Plugins live in `~/.gemini/config/plugins/`, **not** `~/.gemini/antigravity-cli/plugins/` as documented.
- `agy plugin validate` exists and is the fastest conformance check, though it is undocumented.
- Agents must be `agents/<name>/agent.md`; a flat `<name>.md` is silently ignored.
- Remote MCP servers use `serverUrl`, not `url`.
- `gemini-extension.json` also carries `mcpServers`, but `validate` ignores it (`data-agent-kit-plugin` declares 13 servers there and still reports `mcpServers: skipped`), so the plugin declares them in `mcp_config.json`.
- Skills are left unprefixed because they are plugin-scoped — `chrome-devtools-plugin` ships a bare `troubleshooting` skill too.

## Related work

- **#18 (Codex + Gemini CLI)** — different surface (Gemini CLI extensions install to `~/.gemini/extensions/`; this targets `~/.gemini/config/plugins/`), so no runtime collision, but there is real textual overlap in `src/setup.ts`, `README.md`, `.claude/CLAUDE.md` and `src/__tests__/install.test.ts`. Whichever lands second will need a rebase. Both independently arrived at the `~/.cognigy-plugin/config.json` credential fallback.
- **#26 (credentials without `userConfig`)** — touches `src/config.ts` / `src/setup.ts` for the same fallback this relies on; worth landing before or alongside.

- None for existing clients. For anyone who hand-wired a `cognigy` server into `~/.gemini/config/mcp_config.json`, the installer removes that entry and replaces it with the plugin, so only one engine boots.

## Rebased onto main (#18, #26)

`main` gained ChatGPT + Codex, Gemini CLI, `other-hosts`, and the per-client guide structure while this was open. Resolved:

- **Client list unioned** — `antigravity` sits after `gemini`, before the `other-hosts` catch-all. `src/setup.ts` was taken from `main` and this PR's Antigravity blocks re-applied on top, rather than resolving eleven hunks by hand.
- **`runUpdate` ordering** — the Gemini branch returns early, so the Antigravity re-stage is inserted *before* it. Placed after, it would silently never run for anyone without the Gemini extension.
- **`docs/install/antigravity.md` added** — this PR documented Antigravity as a README `<details>` block, which #18's per-client guide structure replaces. Same content, plus the matrix row, the docs link, `--client antigravity` in the flag list, and Antigravity's split update story (engine auto-updates, skills/agents need `cognigy-setup update`) in Staying up to date.

## Additions

- **`agy plugin import gemini` collision.** That command converts installed Gemini CLI extensions into Antigravity plugins. #18 ships a Cognigy *extension* for Gemini CLI, so a user who installed it before [Gemini CLI stopped serving consumer tiers on 18 June 2026](https://developers.googleblog.com/an-important-update-transitioning-gemini-cli-to-antigravity-cli/) and then migrated can end up with an imported copy alongside this plugin — two engines, duplicate tools. `removeLegacyGlobalServer` reconciles a stale global `mcp_config.json` entry but cannot see an imported *plugin*, so the guide explains how to spot and remove it. Worth deciding later whether the installer should detect it.
- **`$schema` on the staged `plugin.json`** — optional per the format docs, recommended, and it gives editors validation when someone opens the staged manifest. Asserted in the manifest test.
- **Fourth archetype in `.claude/skills/add-client-platform/SKILL.md`** — "stage and register", plus the rule this PR demonstrates: driving a client CLI is a choice rather than a default, and `agy plugin install` merging instead of replacing is the case that shows why.

## Verification after the rebase

`tsc --noEmit`, **474 passing** / 3 skipped, `npm run check:manifest`, Prettier — all green.

## Marketplace checklist

- [x] Tool count in `README.md` updated if the tool surface changed — tool surface unchanged (16)
- [x] `README.md` updated if user-facing behavior changed
- [x] Tool annotations/descriptions reviewed if tool behavior changed — no tool behaviour changed
- [x] `LICENSE` / marketplace metadata reviewed if submission requirements are affected — unaffected


PhotonicDev:
## Review round: 4 bugs fixed, 1 simplification, docs corrected

Ran an adversarial review over the diff plus two research passes, then re-ran the installer against a real Antigravity install. Pushed in 132538b.

### Fixed

1. **`enablePluginInConfig` could silently destroy `~/.gemini/config/config.json`.** The read helper swallowed every failure — syntax error, `EACCES`, half-written file — and returned `{}`, after which we wrote our single plugin over the top, discarding the user's other plugins and `userSettings` with no backup. `claudeDesktop.ts` deliberately backs up before its equivalent merge; that contract was unmet here. Now `readJsonForMerge` separates _absent_ from _malformed_, a malformed file is copied to `config.json.bak` before any blank-slate write, and both removal paths refuse to rewrite what they can't parse.

2. **An assetless engine build could wipe a working install and report success.** If `dist/plugin-assets/` were missing, `stagePluginDir` still produced an empty-but-valid plugin; `updateAntigravity` then replaced the installed plugin with it and printed "✓ re-synced (0 skills, 0 agents)". It now throws _before_ anything is removed.

3. **`agy plugin install` merges into an existing plugin dir rather than replacing it.** Found by running the installer twice: files from the previous version survived the upgrade and `agy plugin validate` reported **4 agents for 2 files**. `registerPlugin` now clears any prior install first. This one would not have shown up in unit tests.

4. **`agy` invocation was Windows-broken and failure-blind.** `agy.cmd` cannot be spawned without a shell post-CVE-2024-27980 — the exact hazard `npmRunner.ts` already handles — and a non-zero exit fell through to the fallback while printing "agy not found". Now routed through a helper with the shell/quoting path and a timeout, and a _failing_ `agy` is reported distinctly with its stderr. Also: uninstall no longer claims success when nothing was installed.

### Simplified

Antigravity accepts **flat `agents/<name>.md`** as well as `agents/<name>/agent.md` — verified with `agy plugin validate` — and flat is what other multi-client plugins ship. So our Claude-format agents copy across verbatim and the conversion code is deleted.

### Corrected in docs

**Plugin location is version-dependent.** CLI v1.0.2 relocated installs to `~/.gemini/config/plugins`; older builds staged under `~/.gemini/antigravity-cli/plugins`. Since `agy` decides, `PLUGIN_ROOTS` now lists both — we only _write_ to the former (fallback path) but _look_ in both, or status and uninstall would miss a plugin `agy` put in the other.

### Research notes (informed the design, no action needed)

- **No marketplace exists for Antigravity.** `install <target>` takes a local path or a **git URL**; `plugin@marketplace` and `link` are help-text only, and both `antigravity.google/schemas/v1/*.json` 404. Google's own Data Agent Kit uses a git URL for `agy` and `plugin@marketplace` only for Claude Code/Codex.
- **There is no `userConfig` equivalent.** The `settings`/`envVar` prompt flow is a _Gemini CLI extension_ feature, not Antigravity; `plugin.json` documents only `name` + `description`. Data Agent Kit tells `agy` users to hand-edit `mcp_config.json`. This confirms the credential-file approach here is the only clean option.
- **`agy plugin import claude` is undocumented and lossy.** Judging by the documented `import gemini` path it converts commands→skills and migrates MCP servers, but the sample output _skips agents_ and the docs admit "partial parity". Credentials can't survive translation. Not a viable substitute for this installer.
- **`mcp_config.json` at plugin root is the supported MCP declaration**; `gemini-extension.json` is legacy input to `import gemini` only. `serverUrl` for remote servers is corroborated.

### Possible follow-up (not in this PR)

The established pattern for multi-client plugins — e.g. [EveryInc/compound-engineering-plugin](https://github.com/EveryInc/compound-engineering-plugin) and [bostonaholic/team](https://github.com/bostonaholic/team) — is a **plugin-shaped repo root**: a root `plugin.json` (Antigravity's marker) beside `.claude-plugin/`, sharing one `skills/` dir, so Antigravity users install with `agy plugin install <git-url>` and no installer at all. That would need `skills/` at the repo root rather than under `plugin/`, so it's a layout change worth discussing separately — this PR keeps the installer path, which also handles credentials and the engine launcher that a bare git install cannot.

### Verification after fixes

- `npm test` — 440 passed, 0 failed
- `agy plugin validate` on the live install — 13 skills, 2 agents, 2 mcpServers
- Server booted as Antigravity launches it, no env vars — 16 tools + instructions

dshire (approved):

PR #30, opened 2026-08-19
fix: make auto-updates actually reach users on ChatGPT + Codex and Antigravity

cu4nt0m:
## Summary

Auto-updates were broken on two clients, in two different ways, and the docs asserted update behaviour on both that nobody had verified. Fixed here, with the README stating each client's behaviour in a column instead of a paragraph that has to be reworded every time a client is added.

- **ChatGPT + Codex — frozen by a stale marketplace ref.** Codex *does* auto-update plugins: it re-checks configured Git marketplaces on plugin startup ([openai/codex#17425](https://github.com/openai/codex/pull/17425), default-on since April 2026), comparing `git ls-remote` against the recorded `last_revision` and replacing the cached plugin when the ref has moved. But given a bare `owner/repo`, Codex resolves the ref from the Git checkout the installer runs in — so installing from a clone of this repo on a feature branch pinned the marketplace there. Once that branch merged and stopped moving, auto-upgrade correctly found nothing, forever. Observed live: a ChatGPT install frozen at **1.8.3** while npm and `main` were both at **1.10.0**.
- **Antigravity — skills and agents never updated at all.** Antigravity has no mechanism for this: its [official CLI docs](https://antigravity.google/docs/cli/plugins) list only `agy plugin list | install | enable | disable | uninstall`, `agy plugin install` takes a local directory rather than a remote source, and there is no marketplace or refresh. Plugin files sat untouched from the day they were staged while the engine moved on. The launcher already runs on every boot to pull the latest engine, so it now re-stages skills and agents when the engine version has moved.
- **Docs corrected.** The README and the ChatGPT + Codex guide claimed the plugin runs the engine at `@latest`. It never has — the manifest pins it exactly, which is the point of the plugin/engine version lockstep.
- **Skill updated** so the next client doesn't repeat this. The update lifecycle was the one thing `add-client-platform` said nothing about.

No breaking changes. Codex users on a stale ref are repaired the next time they run the installer. Antigravity users need one `cognigy-setup update` to pick up the new launcher, and are hands-off from then on.

Contains a `feat` commit, so this releases a **minor**.

## Changes

| `src/install/codex.ts` | `buildCodexMarketplaceAddArgs` now passes `--ref main`, so a fresh install can never inherit a branch. New `readCodexMarketplaceRef` reads the ref from the `[marketplaces.cognigy-plugin]` block of `config.toml` (scoped to that one table — the file is the user's and holds unrelated secrets). `installCodex` uses it to drop and re-add a Git registration pinned anywhere other than `main`; a `local` source is left alone, since that is a developer's own wiring. New `buildCodexMarketplaceUpgradeArgs` and `updateCodex` force a refresh now rather than at the next app start. |
| `src/install/desktopLauncher.ts` | After the engine check, the generated launcher imports the engine's own `dist/install/antigravity.js` and calls `updateAntigravity()` when the staged plugin version differs from the engine's — once per release, not every boot. Wrapped in try/catch and gated on a plugin directory existing, so a failure cannot take the MCP server down and Claude Desktop (which shares this launcher) skips it entirely. |
| `src/install/antigravity.ts` | `updateAntigravity` now rewrites the launcher too. It lives outside the versioned engine directory, so it is the one file an engine bump cannot refresh — without this, launcher fixes would never reach an existing install. |
| `src/setup.ts` | `runUpdate` did nothing at all for Codex; it now calls `updateCodex`, gated on the plugin being installed. It also refreshes the Desktop launcher, but only where one already exists, so it cannot create one on a machine using neither client. |
| `README.md` | New **Auto-updates** column on the per-client table — Automatic / Enable once / Manual — replacing the prose paragraph. Also drops the header's claim that only Claude Code pins the engine: Codex and Gemini pin it too; only Claude Desktop and Antigravity float, via the launcher. |
| `docs/install/chatgpt-codex.md` | Replaces the false `@latest` paragraph with how Codex's auto-upgrade actually works, the command to force it early, and a "stuck on an old version?" section pointing at `codex plugin marketplace list`. |
| `docs/install/antigravity.md` | Updating section rewritten: nothing to do, why this is ours rather than Antigravity's, and the one-launch lag before refreshed skills load. |
| `.claude/skills/add-client-platform/SKILL.md` | The update lifecycle was the one thing the skill said nothing about, beyond "a line in `runUpdate()`" — and both bugs here came from that gap. Adds a pre-flight question, an update column to the archetype table so each reference implementation shows how its releases actually travel, and a hard rule with the three questions to answer: does the client refresh the source itself, is anything pinned to something that stops moving, and do plugin files update or only the engine. |
| `src/__tests__/installCodexGemini.test.ts` | Asserts the `--ref main` pin and the upgrade args, extends the "no global mcp server" guard, and covers `readCodexMarketplaceRef` — a healthy `main` pin, the exact stale-branch case that caused this bug, and the null cases (local source, someone else's marketplace, missing file). |
| `src/__tests__/install.test.ts` | Covers the launcher's new step: present, version-gated, ahead of the engine handoff, and failure-tolerant. Since the launcher is a *string* tsc cannot typecheck, one test asserts every `mod.*` symbol it calls is really exported by `antigravity.ts` — so a rename cannot break auto-update silently. |

## Verification

Both paths were exercised for real, not just unit-tested.

- **Codex:** `updateCodex()` returned `{"method":"cli","refreshed":true,"reinstalled":true}` and the ChatGPT app moved from 1.8.3 to 1.10.0.
- **Antigravity:** ran the generated launcher against an isolated `HOME` with a stale plugin staged at 0.0.1. It logged `refreshed Antigravity skills/agents 0.0.1 -> 1.10.0`, rewrote `plugin.json`, staged 13 skills and 2 agents, removed a skill the old version had shipped (replace-not-merge held), and wrote both registries — `import_manifest.json` and the `plugins.cognigy-plugin.enabled` flag. A second run correctly did nothing.

480 tests pass, `tsc` clean, Prettier clean.

## Merge note

PR #27 (Antigravity) touches `docs/install/chatgpt-codex.md` and `README.md` too. Whichever lands second resolves a small conflict in the update sections.

## Test plan

### DIFF
--- a/.claude/skills/add-client-platform/SKILL.md
+++ b/.claude/skills/add-client-platform/SKILL.md
@@ -14,18 +14,19 @@ Verify against the client's **source or official docs**, not blog posts, and che
 1. Does it read an existing manifest of ours as a fallback? If so, what does the fallback *lose* (presentation fields, credential handling)?
 2. Does it interpolate credential placeholders (`${user_config.*}` or equivalent)? Most clients do **not** — see the credential rule below.
 3. Is there a CLI that owns its config file, or must we merge the file ourselves?
+4. **How does the client update an installed plugin?** See the update rule below — this is the question most easily assumed and most often wrong.
 
 Only the parts a fallback can't cover need new code.
 
 ## Pick the archetype
 
-| Archetype | Reference | How it works |
-|---|---|---|
-| Repo-marketplace plugin | `claude-code`, `codex` | Client discovers `plugin/.<client>-plugin/plugin.json` via a marketplace manifest in this repo. Installer drives the client CLI (`marketplace add`), user finishes in the client's plugin UI. |
-| Direct config merge | `claude-desktop` | No CLI. Installer reads/merges the client's own MCP config file, preserving every other key, and chmods it 0600 when it holds secrets. |
-| Packaged extension | `gemini` | A build script emits a release asset; installer runs the client's `extensions install` against the GitHub release. |
-| Stage and register | `antigravity` | The installer builds a plugin directory itself and writes the client's own registry entries — no client CLI involved, works before the client's first launch. |
-| Credentials only | `other-hosts` | The client installs the plugin itself (or takes a hand-written MCP entry); the installer writes `~/.cognigy-plugin/config.json` and wires nothing. |
+| Archetype | Reference | How it works | How updates reach the user |
+|---|---|---|---|
+| Repo-marketplace plugin | `claude-code`, `codex` | Client discovers `plugin/.<client>-plugin/plugin.json` via a marketplace manifest in this repo. Installer drives the client CLI (`marketplace add`), user finishes in the client's plugin UI. | The client re-checks the repo ref and swaps the cached plugin; the new manifest's engine pin pulls the matching engine. **Only works if the ref moves** — pin it to `main` explicitly. |
+| Direct config merge | `claude-desktop` | No CLI. Installer reads/merges the client's own MCP config file, preserving every other key, and chmods it 0600 when it holds secrets. | Nothing in the client updates. The MCP entry points at the launcher, which pulls `@latest` each boot. |
+| Packaged extension | `gemini` | A build script emits a release asset; installer runs the client's `extensions install` against the GitHub release. | The client's own extension auto-update, if it has one — install with that flag on. |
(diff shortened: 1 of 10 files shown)


PhotonicDev (approved):
LGTM

PR #31, opened 2026-08-20
feat(snapshots): add manage_snapshots for backup and rollback

cu4nt0m:
## Summary

- Adds `manage_snapshots` (tool 17) so a client can take a restorable backup before it starts changing an existing agent, and roll the project back on request. Operations: `list` / `create` / `restore` / `delete` / `read_task`.
- Plugin-created backups are identifiable and are the **only** snapshots the plugin will delete; at the snapshot limit `create` creates nothing and asks, then frees a slot by removing the **oldest plugin backup** only.
- `restore` is a no-op preflight report unless `confirm: true`, because restoring deletes every resource in the project before recreating it.

## Changes

| `src/tools/snapshotManagement.ts` | New. Pure logic: marker build/detect, name+timestamp generation, limit evaluation (deletable subset sorted oldest-first), restore preflight body, and the platform failure-reason constants. Mirrors `packageManagement.ts`. |
| `src/tools/definitions.ts` | New `manage_snapshots` entry following the `manage_packages` shape. Description states the project-wide scope and the exclusions up front. |
| `src/schemas/tools.ts` | `manageSnapshotsSchema` as a `z.discriminatedUnion`. `create` takes a short `label`, **not** a `name`, so the marker prefix can never be omitted by the caller. |
| `src/tools/handlers.ts` | `handleManageSnapshots` + dispatch case. Reuses the existing `waitForTask` / `readTask` / `normalizeTask` helpers — snapshot create/restore/delete are the same 202-plus-task pattern as packages. Also adds the once-per-session backup suggestion. |
| `src/tools/filters.ts` | `snapshot` filter exposing the computed `isPluginBackup` (the deletion gate) and dropping download-only noise. `withHints`' constraint loosened from `Record<string, unknown>` to `object` so it accepts a typed interface. |
| `src/instructions.ts` | Capability line + 4 hard rules. **Amends the `delete_resource` rule**, which claimed to be the only way to delete anything — no longer true, so it now names the snapshot exception rather than leaving the model with two contradictory rules. |
| `plugin/skills/snapshot-backups/SKILL.md` | New skill: back up before changing an existing agent, roll back, handle the limit, delete a backup. |
| `plugin/skills/agent-creation/SKILL.md` | One cross-reference bullet at the point that skill starts refining an existing agent. |
| `README.md` | Tool count 16 → 17 and a table row. |
| Tests | 77 new across `tools.test.ts`, `schemas.test.ts`, `filters.test.ts`, and a new `snapshotToolSurface.test.ts`. |

### Design decisions worth reviewing

**Own tool, not new resource types.** `delete_resource` deliberately does **not** learn about snapshots — routing snapshot deletion through it would bypass the plugin-backups-only gate entirely. A test asserts `snapshot` stays out of its enum. This matches the `manage_packages` precedent.

**The marker lives in both `name` and `description`, and both must match.** Snapshots have no metadata field and no PATCH endpoint, so `name` + `description` are the only writable fields — which conveniently makes the marker immutable through the API. Requiring both halves means a hand-made snapshot that happens to share the `[AI Backup] ` prefix is still protected from deletion.

**The backup suggestion is advisory, never blocking.** Modelled on the existing `withRenderSuggestion`, but backed by instance state so "offer once per session" is enforced rather than merely requested in the prompt string. Wired at the dispatcher so all five mutating tools are covered by one rule, and gated on the operation so a read (`manage_flow_nodes get/list/render`) never triggers it. This relies on one server process per session, which holds for the stdio transport the plugin ships.

## Findings from the platform source

These were verified in `service-api` / `service-resources` / `service-ui`, not taken from the docs, and each one changed the implementation:

1. **`:` is a forbidden character in resource names.** The first version generated `HH:mm:ss` timestamps, which would have made *every* `create` fail server-side. Now `HH-mm-ss`. Seconds resolution is required regardless, because snapshot names must be unique per project.
2. **`restore` takes no request body.** The handler resolves the project from the `:snapshotId` path param; sending `projectId` as well raises `"ProjectId was specified in multiple locations"`. The OpenAPI annotation declaring a required `{ projectId }` body is an annotation/implementation mismatch.
3. **The limit rejection never reaches the POST.** `create` returns 202 and the `ConflictError` surfaces on the *task* as `status: "error"`. That `failReason` is translated back into the same `snapshot_limit_reached` shape so the model has one error to understand, not two.
4. **A snapshot attached to an endpoint cannot be deleted at all** — an undocumented `InputOutputError` that surfaces as a 500-class task failure. The rotate path tolerates it and walks to the next-oldest candidate rather than giving up on the first refusal.
5. **The create task carries no `snapshotId`** (it is minted inside the resources controller), so the new snapshot is resolved by name afterwards. `filter` is a substring match on `name` only, which is why marker detection is client-side.
6. **The limit is not readable over the REST API** — `MAX_AMOUNT_SNAPSHOTS_IN_AGENT` is exposed only on service-ui's `/api/config`. Hence `assumedMax`, with the task failReason as the authority.
7. **A snapshot is not a complete backup.** Beyond Endpoints, it excludes **Knowledge AI** (stores, sources, chunks, connectors), Intent Trainer sentences, analytics, contact profiles and logs. Surfaced in the tool description, the `create` response and the restore preflight rather than buried, because a restored RAG agent comes back without its knowledge.

## Test plan

Automated:

- `npm test -- --runInBand` — 519 pass, 77 new
- `npm run check:manifest`
- `npm run build`

Manual — live create + delete round-trip against a real trial project, driven through the real `ToolHandlers`:

- `create` → snapshot accepted with the `HH-mm-ss` name; both markers present verbatim in the stored object; resolve-by-name returned the new id
- `list` → `isPluginBackup: true`, `count`/`atLimit` correct
- `restore` without `confirm` → `applied: false`, no mutating call issued
- `delete` → gated delete succeeded, task reported `done`; project snapshot count returned to 0, tenant left as found
- Confirmed the real payload keys are exactly `_id, description, name, hash, createdAt, createdBy, isPackaged` — the filter matches, and `packageExpiresAt` is absent from the response entirely (declared throughout the platform but never assigned)

**Not exercised: a real destructive `restore`.** Every guard around it is unit-tested and the preflight path is verified live, but no project was actually rolled back. Worth doing on a disposable project before merge.

- None. New tool; no existing tool's inputs or outputs change shape.
- One behavioral note for reviewers: results from `update_ai_agent`, `create_tool`, `update_tool`, `manage_flow_nodes` (mutating operations) and `delete_resource` may now carry an extra `_hints.backupSuggestion` key, once per session. Existing `_hints` values are spread through untouched.

## Marketplace checklist

- [x] Tool count in `README.md` updated if the tool surface changed
- [x] `README.md` updated if user-facing behavior changed
- [x] Tool annotations/descriptions reviewed if tool behavior changed
- [x] `LICENSE` / marketplace metadata reviewed if submission requirements are affected — no change needed


src/tools/handlers.ts
```diff
@@ -4812,6 +4865,541 @@ export class ToolHandlers {
...
+
+        return {
+          operation: "delete",
+          projectId: data.projectId,
+          snapshotId: data.snapshotId,
+          deleted: !timedOut,
+          ...(timedOut ? { pending: true } : {}),
+          taskId,
+          task,
+        };
```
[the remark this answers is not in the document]
cu4nt0m: Correct, and a real bug — fixed in cd01860.

`runSnapshotTask` returns `timedOut: false` on the `waitForCompletion: false` short-circuit, so `deleted: !timedOut` reported success for a task that had only been queued. `delete` now mirrors what `create` and `restore` already did:

```ts
const stillRunning = data.waitForCompletion === false || timedOut;
```

and returns `deleted: false, pending: true` with an action pointing at `read_task`. Regression test added.

src/tools/handlers.ts
```diff
@@ -4812,6 +4865,541 @@ export class ToolHandlers {
...
+    if (this.snapshotCreatedThisSession || this.backupNudgeIssued)
+      return result;
+    this.backupNudgeIssued = true;
+
+    const existing = (result as any)._hints ?? {};
+    (result as any)._hints = {
+      ...existing,
+      backupSuggestion:
+        'This session has changed an existing agent with no backup taken. Offer once, in one short line (do not repeat the offer): ask whether the user wants a restorable backup, and if yes call manage_snapshots { operation: "create", projectId, label: "<why>" }. Mention that a snapshot covers the whole project and does not include Endpoints or Knowledge AI.',
+    };
```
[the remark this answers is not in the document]
cu4nt0m: Correct — fixed in cd01860. This was worse than "unless projectId happens to be in context": the plugin made it impossible.

The agent object does carry the project id, as `projectReference` (verified live against the API), but the `agent` entry in `RESOURCE_FILTERS` stripped it — so `get_resource { resourceType: "agent" }` never returned it and there was no path from an aiAgentId to a projectId through this surface at all.

Fixed at the root rather than in the hint text: the agent filter now exposes `projectId: r.projectReference ?? r.projectId`, and the hint names `get_resource` as where to read it. The existing filter test asserted projectId was stripped — that is exactly the behavior being changed, so it now asserts the field is present while still checking `secret` and `__v` stay out.

src/tools/handlers.ts
```diff
@@ -4812,6 +4865,562 @@ export class ToolHandlers {
...
+        projectId,
+        opts.timeoutMs ?? ToolHandlers.DEFAULT_PACKAGE_TIMEOUT_MS,
+      );
+      return {
+        taskId,
+        task: normalizeTask(task),
+        timedOut,
+        failReason: null,
+      };
+    } catch (error: any) {
```
dshire, line 5245: **[blocker]** This catch conflates two very different outcomes: a platform task that genuinely failed, and a *transient error while polling* (network blip, 5xx after the client's retries on `GET /tasks`). Both come out as `failReason`, and downstream code acts on that as if the operation itself failed:

- on `create`: the model is told "Nothing was backed up, so do not proceed as if a backup exists" while the snapshot may in fact exist → a retry creates a duplicate that eats a limit slot (and `snapshotCreatedThisSession` stays false).
- on `restore`: "the project may be in a partially restored state" for a restore that completed fine.
- worst, in `freeSnapshotSlot`: the candidate is recorded as *skipped* and the loop deletes the **next-oldest backup too** — two irreversible deletions to free one slot.

Suggest distinguishing the cases: let poll/transport errors surface as their own outcome (e.g. `{ pollFailed: true, taskId }`) so callers report "outcome unknown — poll `read_task`" instead of a definitive failure, and so `freeSnapshotSlot` **stops** rather than moving to the next candidate.

src/tools/handlers.ts
```diff
@@ -4812,6 +4865,562 @@ export class ToolHandlers {
...
+  ): Promise<{ deleted: SnapshotSummary | null; skipped: any[] }> {
+    const skipped: any[] = [];
+
+    for (const candidate of candidates) {
+      if (!candidate.id) continue;
+
+      const response = await this.apiClient.delete(
+        `/new/v2.0/snapshots/${candidate.id}`,
+      );
+      const { failReason } = await this.runSnapshotTask(response, projectId, {
```
dshire: **[blocker]** Only `failReason` is destructured here — `timedOut` is dropped. `runSnapshotTask` returns `{ timedOut: true, failReason: null }` when the delete task is still `active` at the deadline, so `!failReason` treats an **unconfirmed** deletion as success: the response claims `deletedToFreeSlot` and the create proceeds against a possibly still-full project.

A timed-out delete should stop the eviction and report `pending` + the `taskId` (poll with `read_task`), not claim the slot was freed.

Related: each candidate delete gets the **full** `opts.timeoutMs` and the create afterwards gets another full budget — no shared deadline, so one call can stack N×timeout. Consider computing one deadline at entry and passing the remaining budget down.

src/tools/handlers.ts
```diff
@@ -4812,6 +4865,562 @@ export class ToolHandlers {
...
+          await this.runSnapshotTask(response, data.projectId, {
+            waitForCompletion: data.waitForCompletion,
+            timeoutMs: data.timeoutMs,
+          });
+
+        // The cap is enforced inside the resources service, so a create that got
+        // past our pre-check (different installation limit, or a race) fails on
+        // the TASK rather than the POST. Translate it into the same shape the
+        // pre-check returns so the model only has one error to understand.
+        if (failReason?.includes(SNAPSHOT_LIMIT_FAIL_REASON)) {
```
dshire, line 5610: **[blocker]** If we get here **after** `freeSnapshotSlot` already deleted a backup (pre-check passed via `confirmDeleteOldest`, but the installation cap is lower than the assumed 10 — the case the comment above describes), this early return drops `freedSlot`/`skippedCandidates`, and `snapshotLimitResult`'s warning says *"No snapshot was created and nothing was deleted"* — which is false: a restorable backup was just permanently destroyed with no trace in the response.

Please thread the eviction outcome into this return (`deletedToFreeSlot`, `skippedCandidates`) and make the warning conditional.

src/tools/handlers.ts
```diff
@@ -4812,6 +4865,562 @@ export class ToolHandlers {
...
+              "This backup covers the WHOLE PROJECT but does NOT include Endpoints or Knowledge AI (stores, sources, chunks). Tell the user that, briefly, if the agent uses knowledge.",
+            action: created
+              ? `To roll back later, call manage_snapshots { operation: "restore", projectId: "${data.projectId}", snapshotId: "${summarizeSnapshot(created).id}" } for a preflight first.`
+              : 'The snapshot was created but could not be resolved by name; use manage_snapshots { operation: "list" } to find its id.',
+          },
+        );
+      }
+
+      case "restore": {
+        const snapshot = await this.readSnapshot(data.snapshotId);
```
dshire, line 5698: **[blocker]** Neither `restore` nor `delete` verifies that this snapshot actually belongs to `data.projectId` — `readSnapshot` fetches by id alone, and (as the create-path comment notes) the platform resolves the project from the snapshot path param. With a stale or copy-pasted `snapshotId` from another project, the confirmed restore **destructively rolls back the wrong project**, while the preflight, warnings, and follow-up hints all name the passed project — and `waitForTask` polls the task with the wrong `projectId` param, so the caller may even be told the restore failed while the other project is being restored unwatched.

The snapshot GET response should carry its project reference — please compare it against `data.projectId` and return a structured error on mismatch (same for `delete`).

src/tools/handlers.ts
```diff
@@ -4812,6 +4865,562 @@ export class ToolHandlers {
...
+
+    return { deleted: null, skipped };
+  }
+
+  async handleManageSnapshots(args: any): Promise<any> {
+    const data = schemas.manageSnapshotsSchema.parse(args);
+
+    switch (data.operation) {
+      case "list": {
+        const items = await this.listSnapshots(data.projectId, {
```
dshire: **[blocker]** Caller pagination flows straight into the limit evaluation: `evaluateSnapshotLimit` counts whatever page it is handed, so `list` with `limit: 5` on a 10-snapshot project returns `count: 5, atLimit: false` — the model believes there's room and skips the free-a-slot conversation. With `skip > 0` it's worse: `oldestDeletableBackup` is picked from a page that may not contain the oldest plugin backup at all, and that's the snapshot we show the user as the eviction candidate ("deletion preference is the oldest" is the agreed rule).

Suggest what the create pre-check already does: compute `count`/`atLimit`/`oldestDeletableBackup` from an unpaginated fetch, and apply the caller's `limit`/`skip` only to the returned `snapshots` array.

src/tools/handlers.ts
```diff
@@ -4871,9 +5480,19 @@ export class ToolHandlers {
...
+          result = await this.handleManageSnapshots(args);
+          break;
         default:
           throw new Error(`Unknown tool: ${toolName}`);
       }
+      // Offer a backup the first time this session changes an EXISTING agent.
+      // Done here rather than inside each handler so every return path in all
+      // five tools is covered by one rule. create_ai_agent is deliberately
+      // absent: creating is additive, so there is nothing to roll back to.
+      if (ToolHandlers.isBackupWorthyCall(toolName, args) && result) {
```
dshire: **[should-fix]** `result` being truthy isn't the same as "something mutated": several of these handlers return structured `{ error: ... }` objects on validation failure instead of throwing (e.g. `manage_flow_nodes create` without `nodeType`, `delete_resource` for a tool without `aiAgentId`). That burns the once-per-session nudge on a no-op — `backupNudgeIssued` is set, the hint lands on an error payload, and the *real* first mutation never gets an offer.

Suggest skipping the nudge when the result carries an `error` key (or only attaching it on confirmed-success paths).

src/tools/handlers.ts
```diff
@@ -4812,6 +4865,562 @@ export class ToolHandlers {
...
+  /**
+   * Attach a one-time offer to back the project up before it is changed further.
+   * Same shape and spirit as withRenderSuggestion: merged into `_hints` under
+   * its own key so it can never clobber a real warning, and worded so the model
+   * asks once instead of nagging. Unlike that helper this one is backed by an
+   * instance flag, so "once per session" is enforced rather than merely
+   * requested. Purely advisory — it never blocks a mutation and costs no API
+   * call.
+   */
+  private withBackupSuggestion<T extends object>(result: T): T {
```
dshire: **[should-fix]** Two blind spots in the session state here:

1. **Session-created resources.** Nothing records what this session created, so after `create_ai_agent` (correctly nudge-free), the follow-up `create_tool`/`update_ai_agent` on that *brand-new* agent fires a hint claiming "This session has changed an existing agent" — exactly the case the skill text excludes ("not when creating a new agent"). Suggest remembering ids created this session (agent/project on `create_ai_agent` success is the big one) and suppressing the nudge for calls targeting them.

2. **Declines are invisible.** In a skill-capable client the model asks *before* the first change per the instructions; if the user declines, the model proceeds — and this post-mutation hint then tells it to ask again: the exact double-ask the offer-once rule exists to prevent. Cheap fix: suppress the nudge once the session has made **any** `manage_snapshots` call (even `list`) — evidence the backup conversation already happened.

src/tools/handlers.ts
```diff
@@ -571,6 +584,46 @@ export class ToolHandlers {
   private static readonly DEFAULT_PACKAGE_TIMEOUT_MS = 600000;
   private static readonly TASK_POLL_INTERVAL_MS = 3000;
 
+  /**
+   * Tools that change an EXISTING agent, and so warrant a backup offer.
+   * create_ai_agent is deliberately absent: creating is additive, so there is
+   * nothing to roll back to.
+   */
+  private static readonly BACKUP_WORTHY_TOOLS = new Set([
```
dshire: **[should-fix]** Two gaps in this set, and a rot risk:

- `audit_voice_agent` with `apply: true` PATCHes existing node configs and prepends flow nodes — the bulkiest single mutation the plugin can make to an existing agent, with no backup offer. Please add it, gated on `apply` (the dry-run stays nudge-free).
- `manage_settings` mutates project settings, which **are** snapshot-covered — worth considering too. (`setup_llm` / package import are additive, so excluding them matches the `create_ai_agent` rationale.)
- Nothing forces tool 18 to be classified — the list will silently rot. A surface test asserting every tool is either in this set, read-only-annotated, or explicitly exempted would catch that (or derive the set from the `readOnlyHint`/`destructiveHint` annotations in definitions.ts).

src/tools/snapshotManagement.ts
```diff
@@ -0,0 +1,207 @@
...
+ * Snapshots have no metadata field and no PATCH endpoint — `name` and
+ * `description` are the only writable fields, and they are immutable once
+ * written. So the marker that distinguishes a plugin-created backup from a
+ * human-created snapshot has to live in both of them.
+ *
+ * Requiring BOTH halves to match means a human snapshot that happens to share
+ * the name prefix is still protected from plugin deletion.
+ */
+export const AUTO_BACKUP_NAME_PREFIX = "[AI Backup] ";
+export const AUTO_BACKUP_MARKER = "cognigy-plugin:auto-backup:v1";
```
dshire: **[should-fix]** Please drop the `:v1` — write and match plain `cognigy-plugin:auto-backup`. Descriptions are immutable, and `isAutoBackup` matches this exact literal: if the marker is ever bumped to `:v2`, every existing backup instantly stops being `isPluginBackup` — undeletable via the tool and invisible to `evaluateSnapshotLimit`'s deletable set, so a project full of old plugin backups dead-ends at the limit. We don't need format versioning here (decided with David); if you'd rather keep the suffix, then `isAutoBackup` must match the stable prefix `cognigy-plugin:auto-backup:` instead of the full literal. No releases exist yet, so changing the literal is free right now.

src/tools/snapshotManagement.ts
```diff
@@ -0,0 +1,207 @@
...
+    `${p2(now.getHours())}-${p2(now.getMinutes())}-${p2(now.getSeconds())}`
+  );
+}
+
+/**
+ * Build the name+description for a plugin backup. The caller supplies only a
+ * short label — the plugin owns the rest so the marker can never be omitted.
+ * `now` is injected so tests are deterministic.
+ */
+export function buildAutoBackupFields(
```
dshire, line 98: **[should-fix]** Agreed design addition (David): backup names should carry a **monotonic version number** — `[AI Backup] v<N> <label> — <timestamp>` — so backups are addressable ("restore v3") instead of timestamp-only. N = max version parsed from the project's existing marked snapshots + 1; never reuse numbers after deletion so references stay unambiguous forever. The create path already fetches the full snapshot list for the limit pre-check, so the version can be derived from that same response — no extra API call. Suggest exposing the parsed `version` on `SnapshotSummary` too, so `list` output shows it.

src/tools/snapshotManagement.ts
```diff
@@ -0,0 +1,207 @@
...
+
+/**
+ * Snapshot names must be unique within the project, so the timestamp carries
+ * SECONDS — a minute-resolution stamp lets two backups in the same minute
+ * collide with a 409 that only surfaces on the task, well after the fact.
+ *
+ * The time uses `HH-mm-ss`, NOT `HH:mm:ss`: isValidResourceName rejects `:`
+ * outright, so a colon here would make every single create fail server-side.
+ */
+export function formatBackupTimestamp(now: Date): string {
```
dshire, line 86: **[nit]** This stamps **local** time while package names use UTC (`buildExportPackageName` via `toISOString`), so the two naming schemes sort inconsistently across timezones — and two sessions in different timezones (or the same local second) can generate byte-identical names, surfacing only as a generic task `failReason` 409. Suggest UTC here for consistency; full consolidation with the package helper can be the follow-up PR.

src/tools/handlers.ts
```diff
@@ -4812,6 +4865,562 @@ export class ToolHandlers {
...
+
+        // The create task does not carry the new snapshot's id — it is minted
+        // inside the resources controller — so resolve it by name. `filter` is a
+        // substring match on name, and the generated name carries a
+        // second-resolution timestamp, so this is unambiguous.
+        const matches = await this.listSnapshots(data.projectId, {
+          filter: fields.name,
+        });
+        const created =
+          matches.find((s: any) => s?.name === fields.name) ?? matches[0];
```
dshire: **[should-fix]** The `?? matches[0]` fallback can bind the **wrong snapshot's id** into the response — and that id flows straight into the `_hints` action recommending a future restore, i.e. a wrong point-in-time restore that wipes the very changes the backup was taken to protect. If the exact `===` match fails (server-side whitespace/em-dash normalization, fuzzy filter semantics — also untested: the generated name starts with regex metachars `[AI Backup] `), returning the null path is strictly safer: the code already has a good fallback message ("use list to find its id"). Suggest dropping `?? matches[0]`.

src/tools/definitions.ts
```diff
@@ -1714,4 +1714,67 @@ After creating, use talk_to_agent to test.`,
...
+    annotations: {
+      title: "Manage Snapshots",
+      readOnlyHint: false,
+      destructiveHint: true,
+      idempotentHint: false,
+      openWorldHint: true,
+    },
+    inputSchema: {
```
dshire, line 1730: **[should-fix]** This JSON schema and `manageSnapshotsSchema` (zod) disagree:

- zod's `list` accepts `limit`/`skip` (paginationSchema) — not declared here, so models never learn they exist and strict-validation clients reject them.
- zod bounds `timeoutMs` (int 1000–3600000) and `label` (max 120) — declared unbounded here, so JSON-legal values like `timeoutMs: 500` or a 200-char label die as raw ZodErrors instead of the tool's structured error shapes.

Please sync the two contracts field-for-field.

src/tools/filters.ts
```diff
@@ -116,6 +122,19 @@ export const RESOURCE_FILTERS: Record<string, (raw: any) => any> = {
     name: r.name ?? r.label,
     toolType: r.toolType ?? r.type,
   }),
+  // `isPluginBackup` is the deletion gate, so it belongs in every snapshot view.
+  // `hash` and `packageExpiresAt` are dropped: both only matter for downloading,
+  // which this plugin deliberately does not do (and the platform never actually
+  // assigns packageExpiresAt).
+  snapshot: (r) => ({
```
dshire: **[should-fix]** This is a field-for-field re-implementation of `summarizeSnapshot` (snapshotManagement.ts) — `list` renders snapshots through this projection while `create`/`restore`/`delete` use `summarizeSnapshot`, and both independently compute `isPluginBackup`, the flag that gates deletion. Two parallel code paths for a safety-critical flag will drift. Suggest delegating: `snapshot: summarizeSnapshot` (import it here instead of `isAutoBackup`).

src/instructions.ts
```diff
@@ -29,10 +30,14 @@ RULES:
 - Cognigy Connections are PROJECT-SCOPED: a connectionId from project A cannot be used in project B (fails with "Connection does not exist"). The ONLY way to share one across projects is package export/import. Never pass a cross-project connectionId to setup_llm, and never use dangerouslySkipConnectionTest to bypass a missing or cross-project connection.
 - For knowledge / RAG: the embedding model and the project-level Knowledge Search model are distinct; call set_knowledge_ai BEFORE creating the store; attach knowledge as a tool (not the persona) by default.
 - talk_to_agent hits a DIFFERENT base URL (endpoint-*.cognigy.ai) — not the API base URL.
-- delete_resource is the ONLY way to delete anything.
+- delete_resource is the ONLY way to delete anything, with one exception: snapshots are deleted via manage_snapshots, and only the plugin's own "[AI Backup]" snapshots can be deleted at all.
```
dshire: **[should-fix]** Two accuracy issues in the strengthened rule:

1. "with **one** exception" is factually wrong — `manage_flow_nodes` has had `operation: "delete"` all along (and this PR's own `READ_ONLY_OPERATIONS` logic treats it as a mutation). In skills-less clients this always-on enumeration is the only guidance, and it steers a node-deletion request into `delete_resource`, which has no node resourceType.
2. "only the plugin's own `[AI Backup]` snapshots can be deleted" describes the gate by name prefix alone — the real gate needs the name prefix **and** the description marker (the tool description gets this right: "both markers"). A model in a skills-less client will promise deletion of a human snapshot named `[AI Backup] …` and hit `not_a_plugin_backup` with no rule explaining why.

src/instructions.ts
```diff
@@ -29,10 +30,14 @@ RULES:
...
 - audit_voice_agent runs a dry-run first; apply fixes only after review.
-- manage_webchat creates a new endpoint when endpointId is omitted, and updates an existing one only when endpointId is provided (find it via list_resources { resourceType: "endpoint", projectId }). It ALWAYS returns demoWebchatUrl — present it to the user every time as a clickable link; do NOT tell the user to find the URL in the Cognigy UI.`;
+- manage_webchat creates a new endpoint when endpointId is omitted, and updates an existing one only when endpointId is provided (find it via list_resources { resourceType: "endpoint", projectId }). It ALWAYS returns demoWebchatUrl — present it to the user every time as a clickable link; do NOT tell the user to find the URL in the Cognigy UI.
+- BEFORE the first change to an EXISTING agent in a session, ask the user once whether they want a restorable backup; if yes, call manage_snapshots { operation: "create", projectId, label } and wait for it to finish before changing anything. Do not ask again, and do not ask when creating a new agent.
```
dshire: **[nit]** These four bullets are step-by-step single-tool workflow (restore preflight sequence, full limit-recovery procedure) — per the repo's CLAUDE.md, instructions.ts should stay terse cross-tool hard rules while skills own workflow detail. The same procedures already live in the SKILL.md, the tool description, *and* the runtime `_hints` (which arrive exactly when relevant). Suggest collapsing to ~2 lines: offer-once-before-first-change, and restore-is-destructive/preflight-then-confirm. Every session in every client pays for these tokens, snapshot-related or not.

plugin/skills/snapshot-backups/SKILL.md
```diff
@@ -0,0 +1,174 @@
...
+
+`label` is a short reason, not a name. The plugin builds the name itself as
+`[AI Backup] <label> — <timestamp>` so the backup is always identifiable and always
+uniquely named.
+
+### Roll back to a backup
+
+1. `manage_snapshots { operation: "list", projectId: "<projectId>" }`
+2. Pick the snapshot. `isPluginBackup: true` marks the ones this plugin created.
+3. Get a preflight — this changes **nothing**:
```
dshire: **[should-fix]** One nuance agreed with David: when the restore target is **not** the backup this session just created (an older or human-made snapshot), the current project state is about to be destroyed with no way back — offer to create a safety snapshot of the current state first, if a slot is free. When the restore merely reverts this session's own changes back to its own fresh backup, skip the offer (redundant). One bullet between steps 2 and 3 covers it.

src/tools/handlers.ts
```diff
@@ -4812,6 +4865,562 @@ export class ToolHandlers {
...
+        ...(failReason.includes(SNAPSHOT_IN_USE_FAIL_REASON)
+          ? { inUseByEndpoint: true }
+          : {}),
+      });
+    }
+
+    return { deleted: null, skipped };
+  }
+
+  async handleManageSnapshots(args: any): Promise<any> {
```
dshire, line 5460: **[follow-up, non-blocking]** Noting for a future PR rather than this one: the task plumbing added here (`read_task` case, `runSnapshotTask`'s taskId extraction/guard) mirrors the manage_packages equivalents almost verbatim, and `snapshotTimeoutSchema` is the 4th copy of the same timeout bounds. Consolidating both (plus the timestamp/sanitizer helpers, see the snapshotManagement.ts comments) would give packages and snapshots one shared async-task toolkit instead of two drifting ones.

dshire (changes requested):
Really solid work overall — the platform-source verification (colon-free names, bodyless restore, limit errors surfacing on the task, endpoint-attached snapshots refusing deletion) is exactly the diligence this feature needs, and the dual-marker + preflight/confirm safety architecture matches what we agreed in the meeting.

Requesting changes because the worst findings cluster where the feature must never be wrong: the **limit-eviction path can destroy backups and then misreport it** (poll errors conflated with task failures → can delete two backups; timed-out deletes counted as done; one path answers "nothing was deleted" after a deletion), and **restore never checks the snapshot belongs to the passed project**.

**Blockers** are marked inline: error/timeout handling in `runSnapshotTask`/`freeSnapshotSlot`, the freed-slot misreport, the restore ownership check, and `list`'s paginated miscount.

**Also requested** (inline): nudge fixes (don't burn it on failed calls, track session-created resources, cover `audit_voice_agent apply:true`), drop the `:v1` from the marker + prefix-match, add version numbers to backup names (`[AI Backup] v<N> …`), sync the JSON inputSchema with zod, drop the `matches[0]` fallback, and have the `snapshot` filter delegate to `summarizeSnapshot`.

**Not inline-able:** `.claude/CLAUDE.md` still says "16 tools" — please bump to 17 alongside the README change.

**Fine as follow-up PRs** (suggestion, not blocking this one): consolidate the duplicated task-timeout zod schema (now 4 copies across packages + snapshots) and the timestamp/name-sanitizer helpers (`formatBackupTimestamp`/`FORBIDDEN_NAME_CHARS` vs `buildExportPackageName`/`sanitizePackageFileName`) — the two sanitizer character classes already disagree (control chars vs `¥`).

cu4nt0m:
## Update: requirement 1 reworked (5a17e5b)

Testing in a real session surfaced a design flaw in the backup reminder, so it has been replaced.

**What was wrong.** The reminder was an advisory `_hints.backupSuggestion` attached to the **result** of a mutating call. By construction that cannot arrive before the mutation — so Claude rewrote an agent's persona and *then* offered the backup. A snapshot taken at that point captures the already-changed state, which is no rollback point at all. It was worse than absent: it gave the model a natural-looking place to raise the backup late, so the mechanism actively produced the wrong order.

**What replaces it.** A pre-flight gate. The first attempt to change an existing agent in a session performs no call and returns `error: "backup_not_offered"`; the client asks the user, calls `manage_snapshots { operation: "create" }` or the new `decline` operation, then retries. Same refuse-report-retry idiom already used for the restore preflight and `snapshot_limit_reached`, so no new parameters on any existing tool.

**Deliberate limits**, each with a test:

- Trips **at most once per session**. A gate that held out for compliance could deadlock an automated subagent; one forced pause is what "ask at the start of the session" needs.
- **Never fires for a resource this session created.** `cognigy-agent-builder` refines an agent immediately after creating it and there is no prior state to roll back to, so `create_ai_agent` now records the agent and flow ids it mints.
- Also gates `audit_voice_agent`, but **only** with `apply: true` — its default dry-run mutates nothing. This closed a hole: apply mode rewrites an existing agent's nodes and previously bypassed the gate entirely. `cognigy-voice-go-live` is told how to handle it.
- A **queued** create does not release the gate. An unfinished backup is not a backup.

`manage_settings` is intentionally left ungated — it changes project-level voice and Knowledge AI settings, and Knowledge AI is not in a snapshot anyway.

**Blast radius, stated plainly:** the gate holds the first mutating call in every fresh session, which broke 77 existing test assertions. Suites that exercise other tools now answer the gate up front, the way a real session does once the user has replied. That is the cost of making requirement 1 actually hold.

Verified live against a real project: the first `update_ai_agent` on a pre-existing agent was held with `changed: false` and no API call issued; `decline` released it; create/preflight/delete round-trip still leaves the tenant at its original snapshot count. 529 tests.

src/tools/handlers.ts
cu4nt0m: Fixed in 3572890. `waitForTask` now throws `TaskFailedError` only when the platform reported the task as failed/cancelled; anything else (transport error, unreadable task) comes back as `pollFailed`, and create/restore/delete return `task_status_unknown` with `outcomeUnknown: true` plus a `read_task` instruction instead of a definitive failure. `freeSnapshotSlot` stops on it rather than moving to the next candidate.

src/tools/handlers.ts
cu4nt0m: Fixed in 3572890. `timedOut` is destructured and halts the eviction, reported as `eviction_incomplete` with the `taskId` to poll — no claimed `deletedToFreeSlot`, no create. The budget point is handled too: one deadline is computed at entry and every candidate plus the create draws from the remaining time.

src/tools/handlers.ts
cu4nt0m: Fixed in 3572890. `snapshotLimitResult` now takes the eviction outcome, so `deletedToFreeSlot` and `skippedCandidates` appear in the response and the warning says the backup was already permanently deleted instead of claiming nothing was.

src/tools/handlers.ts
cu4nt0m: Fixed in 3572890. Both operations now verify the snapshot is in the passed project and return `snapshot_project_mismatch` without acting. One correction on the suggested fix: the snapshot object carries no project reference to compare against — `GET /v2.0/snapshots/{snapshotId}` exposes only name/description/isPackaged/_id/hash/createdBy/createdAt — so ownership is established by membership in the project's snapshot list, which needs no undocumented field.

src/tools/handlers.ts
cu4nt0m: Fixed in 3572890. `count`, `atLimit` and `oldestDeletableBackup` are computed from an unpaginated fetch; the caller's `limit`/`skip` only slice the returned `snapshots` array.

src/tools/handlers.ts
cu4nt0m: Already addressed by 5a17e5b, which landed after this review: the gate now holds the call before the handler runs, so there is no post-hoc hint to attach to an `{ error }` payload and the offer cannot be burned on a no-op.

src/tools/handlers.ts
cu4nt0m: Already addressed by 5a17e5b: `resourcesCreatedThisSession` records ids created this session and suppresses the gate for them, and a `decline` operation records the user saying no so the gate stops holding calls.

src/tools/handlers.ts
cu4nt0m: Fixed in 3572890. `audit_voice_agent` (gated on `apply: true`) and `manage_settings` are both in `BACKUP_WORTHY_TOOLS`, which is now exported so a surface test can assert every tool is read-only-annotated, backup-worthy, or listed as an explicit exemption with a reason — tool 18 fails that test until someone classifies it.

src/tools/snapshotManagement.ts
cu4nt0m: Fixed in 3572890. The marker is plain `cognigy-plugin:auto-backup` and is matched as a substring, so anything already written with the `:v1` suffix still reads as a plugin backup. Test added for exactly that case.

src/tools/snapshotManagement.ts
cu4nt0m: Fixed in 3572890. Names are `[AI Backup] v<N> <label> — <timestamp>`, with N derived from the snapshot list the create path already fetches for the pre-check, plus a session floor so deleting the newest backup cannot hand the same number to a second snapshot. `SnapshotSummary` exposes `version`, so `list` shows it.

src/tools/snapshotManagement.ts
cu4nt0m: Fixed in 1950177. `formatBackupTimestamp` stamps UTC now, so it sorts consistently with the package names and two timezones cannot mint the same name. Full consolidation with the package helper is left for the follow-up PR as suggested.

src/tools/handlers.ts
cu4nt0m: Fixed in 3572890. The `?? matches[0]` fallback is gone — only an exact name match binds an id, and the null path keeps the existing "use list to find its id" message.

src/tools/definitions.ts
cu4nt0m: Fixed in 3572890. The JSON schema now declares `limit` (1-100) and `skip`, and bounds `timeoutMs` (integer, 1000-3600000) and `label` (1-120). A surface test asserts the bounds and round-trips boundary values through `manageSnapshotsSchema` so the two contracts cannot drift apart silently.

src/tools/filters.ts
cu4nt0m: Fixed in 3572890. The filter is now `snapshot: summarizeSnapshot`, so `isPluginBackup` is computed in exactly one place.

src/instructions.ts
cu4nt0m: Fixed in 3572890. The rule now names `manage_flow_nodes { operation: "delete" }` as the other non-`delete_resource` deletion path, and describes the gate as the name prefix AND the description marker, naming the `not_a_plugin_backup` error a model will otherwise hit unexplained.

src/instructions.ts
cu4nt0m: Fixed in 1950177. Collapsed to two bullets: the backup gate, and restore being destructive behind a preflight. The scope caveat (Endpoints and Knowledge AI not covered) is folded into the restore line rather than kept as its own bullet; the limit-recovery procedure is left to the tool description, the skill and the runtime hints.

plugin/skills/snapshot-backups/SKILL.md
cu4nt0m: Fixed in 3572890. Added as step 3 of "Roll back to a backup": offer a safety snapshot of the current state when the target is an older or human-made snapshot, and skip it when the restore just reverts this session's own changes to its own fresh backup.

src/tools/handlers.ts
cu4nt0m: Not done in this PR — leaving this open as the follow-up you suggested. Consolidating the task plumbing (read_task case, taskId extraction, the four copies of the timeout bounds) plus the timestamp/sanitizer helpers is a separate change, and the sanitizer character classes disagreeing (control chars vs `¥`) belongs in the same pass.

cu4nt0m:
Thanks — all five blockers and every should-fix are addressed, pushed as `3572890` and `1950177`. Inline threads are answered and resolved individually; short summary:

**Blockers**
- **Poll error read as task failure** (`runSnapshotTask`) — `waitForTask` now throws `TaskFailedError` only for a platform-reported failure; anything else surfaces as `pollFailed`, and create/restore/delete return `task_status_unknown` / `outcomeUnknown: true` with a `read_task` instruction.
- **Timed-out delete counted as a freed slot** — `freeSnapshotSlot` halts on `timedOut` or `pollFailed` instead of walking to the next candidate, so one slot can never cost two backups. All eviction attempts plus the create now share one deadline.
- **Freed-slot misreport** — the eviction outcome is threaded into the limit result; the warning no longer claims nothing was deleted after a backup was destroyed.
- **No project ownership check** — restore and delete verify the snapshot belongs to the passed project and return `snapshot_project_mismatch` without acting. Ownership is checked by membership in the project's snapshot list: the snapshot object carries no project reference (the REST schema exposes only name/description/isPackaged/_id/hash/createdBy/createdAt).
- **Paginated miscount in `list`** — count/atLimit/oldestDeletableBackup come from an unpaginated fetch; `limit`/`skip` only slice the returned array.

**Should-fix**
- `manage_settings` and `audit_voice_agent` (gated on `apply: true`) are backup-worthy; the set is exported and a surface test forces every tool to be read-only, backup-worthy, or exempt with a stated reason.
- Marker is plain `cognigy-plugin:auto-backup`, matched as a substring, so `:v1` descriptions still register.
- Backup names carry a monotonic `v<N>`, never reused within a project; `version` is exposed on summaries.
- `?? matches[0]` removed — exact name match only.
- JSON input schema synced with zod (`limit`/`skip`, `timeoutMs` bounds, `label` bounds), asserted by a test.
- Snapshot filter delegates to `summarizeSnapshot`.
- `instructions.ts` no longer claims snapshots are the only exception to `delete_resource`, and describes the gate as both markers.
- The skill offers a safety snapshot before restoring anything other than this session's own backup.
- `.claude/CLAUDE.md` now says 17 tools.

The two nudge findings (burning the offer on an error payload, and session-created resources / declines) were already fixed by `5a17e5b`, which landed after your review — noted on those threads.

**Nits** — timestamps are UTC now, and the four snapshot bullets in `instructions.ts` are collapsed to two.

Left open deliberately: the non-blocking follow-up thread about consolidating the async-task plumbing, the four copies of the timeout schema, and the timestamp/sanitizer helpers. That's a separate PR.

New tests cover each fix: poll-failure vs real failure, halted eviction (asserting exactly one delete), timed-out eviction, the reported destroyed backup, cross-project restore and delete, paginated list counting, version numbering and the no-reuse floor, exact-match resolution, and the schema/classification surface checks. 553 tests pass, `tsc` clean.

src/tools/handlers.ts
```diff
@@ -4812,6 +4930,908 @@ export class ToolHandlers {
...
+          deleted: true,
+          taskId,
+          task,
+        };
+      }
+
+      // Records that the user was asked and said no, so the backup gate stops
+      // holding calls for the rest of the session. Touches no API.
+      case "decline": {
+        this.backupDeclinedThisSession = true;
```
dshire: **[fix before merge]** `decline` takes a required `projectId` but ignores it — this flag (and `snapshotCreatedThisSession`) is session-global, so declining a backup on project A releases the gate for **every** project touched later in the session. Concrete risk: user declines on a sandbox ("it's a throwaway"), then switches to a production project — the first destructive change to production sails through with no offer.

Suggest keying the gate state per project: `backupDeclinedForProject: Set<string>` and `snapshotCreatedForProject: Set<string>`. The wrinkle is that the gate fires on calls that don't carry a projectId (`update_ai_agent { aiAgentId }`), so exact scoping needs a resolve step — a pragmatic middle ground is to keep the trip global but scope *decline* and *created* by the projectId they were called with, and re-arm the gate when a held-worthy call arrives for a project that has neither declined nor snapshotted.

src/tools/handlers.ts
```diff
@@ -1403,6 +1516,11 @@ export class ToolHandlers {
         ? "configured"
         : "unknown";
 
+      // Remember what this session minted so the backup gate leaves it alone.
+      for (const id of [agentId, agent.referenceId, flowId, flow.referenceId]) {
```
dshire: **[fix before merge]** When this create also created the **project** (`createdProject` is true above), the projectId never enters `resourcesCreatedThisSession` — and `targetsNewResource` only checks `aiAgentId`/`flowId`/`id`, never `projectId`. So the gate holds mutations on brand-new projects, contradicting its own exemption rationale ("changes to brand-new material are additive").

Concrete case: `create_ai_agent` (new project) → `manage_settings { operation: "set_knowledge_ai", projectId }` — a normal next step of the knowledge flow — gets held with "first change to an existing agent", and if the user says yes, a pointless snapshot of a seconds-old project eats one of its ~10 slots. Fix: add the projectId to the set when `createdProject` is true, and add `projectId` to the keys `targetsNewResource` checks (only matching against ids that were actually recorded, so an *existing* project passed by a later call can't false-match).

src/tools/handlers.ts
```diff
@@ -4812,6 +4930,908 @@ export class ToolHandlers {
...
+          description: fields.description,
+        });
+
+        const { taskId, task, timedOut, failReason, pollFailed, pollError } =
+          await this.runSnapshotTask(response, data.projectId, {
+            waitForCompletion: data.waitForCompletion,
+            timeoutMs: ToolHandlers.remainingMs(deadline),
+          });
+
+        if (pollFailed) {
```
dshire, line 5588: **[fix before merge]** This path returns `created: false` **and** `outcomeUnknown: true` in the same object. `created: false` is the key every other create outcome uses authoritatively to mean "no backup exists, safe to retry" — a caller keying on the boolean rather than the flag/warning will retry and mint the duplicate snapshot this redesign exists to prevent. Suggest `created: null` (or omitting the key) on the unknown path, so the boolean never lies; the schema-parity test can pin it.

src/tools/handlers.ts
```diff
@@ -556,6 +570,42 @@ async function resolveFlowForAgent(
...
+ * package import. Exported so a surface test can assert every tool is
+ * classified — a new tool 18 must land here, be read-only, or be listed as an
+ * explicit exemption.
+ */
+export const BACKUP_WORTHY_TOOLS = new Set([
+  "update_ai_agent",
+  "create_tool",
+  "update_tool",
+  "manage_flow_nodes",
+  "delete_resource",
```
dshire, line 595: **[author's choice]** `delete_resource` is gated for every resourceType, including `endpoint` and `knowledge_store` — both in `SNAPSHOT_EXCLUSIONS`, i.e. things a restore cannot bring back — while the surface test exempts `manage_webchat`/`manage_knowledge` for exactly that reason. The offer text does disclose the exclusions, so this over-holds rather than falsely promises, but the classification is inconsistent: deleting an endpoint via `delete_resource` is held, deleting it via a webchat flow never would be. Options: skip the gate for resourceTypes in the exclusions list, or keep it and have the offer explicitly say "a snapshot will NOT protect this deletion" for those types.

src/tools/handlers.ts
```diff
@@ -4821,6 +5841,20 @@ export class ToolHandlers {
     });
 
     try {
+      // Before anything mutates an existing agent, give the user one chance to
+      // take a backup. Must precede the switch — a post-hoc hint arrives after
+      // the change and is therefore useless.
+      if (
+        ToolHandlers.isBackupWorthyCall(toolName, args) &&
+        !this.targetsNewResource(args)
```
dshire, line 5985: **[author's choice]** The gate runs before each handler's zod validation, so an *invalid* first mutating call consumes the one-shot trip (and the caller sees `backup_not_offered` instead of the validation error, which only surfaces on the post-answer retry). Combined with the deliberate trip-once design, a client that fixes its args and blindly retries proceeds unprotected without the user ever being asked about a call that could actually run. If you want to keep trip-once (reasonable), an alternative is to not set `backupGateTripped` until a call *passes its schema* — e.g. run a lightweight parse for the gated tools first, or reset the trip when the retried call fails validation.

src/tools/snapshotManagement.ts
```diff
@@ -0,0 +1,245 @@
...
+ * creating another cannot hand the same number to two different snapshots.
+ */
+export function nextBackupVersion(snapshots: any[], floor = 0): number {
+  const highest = snapshots
+    .filter(isAutoBackup)
+    .reduce((max, s) => Math.max(max, parseBackupVersion(s?.name) ?? 0), 0);
+  return Math.max(highest, floor) + 1;
+}
+
+export function summarizeSnapshot(raw: any): SnapshotSummary {
```
dshire, line 161: **[author's choice]** `version` is parsed from the name with no `isAutoBackup` guard, so a human snapshot named `[AI Backup] v9 test` reports `version: 9` next to `isPluginBackup: false` — contradictory fields, and the model may offer "restore v9" as if it belonged to the plugin's numbering. (`nextBackupVersion` filters to marked backups first, so the counter itself is safe.) One-line fix: `version: isAutoBackup(raw) ? parseBackupVersion(raw?.name) : null`.

src/tools/definitions.ts
```diff
@@ -1714,4 +1714,84 @@ After creating, use talk_to_agent to test.`,
...
+          type: "string",
+          description:
+            "24-char hex snapshot ID. Required for restore and delete. Must belong to projectId.",
+        },
+        limit: {
+          type: "integer",
+          minimum: 1,
+          maximum: 100,
+          description:
+            "For list: how many snapshots to return (default 100). The reported count, atLimit and oldestDeletableBackup always describe the whole project, never just this page.",
```
dshire, line 1753: **[author's choice]** "default 100" is stale: with `limit` omitted, the handler now returns *every* snapshot (`all.slice(skip, undefined)` after the pagination-following fetch — up to 20×100). Low impact at the platform's ~10 cap, but installations can raise it, and then an omitted limit floods the tool result with full snapshot objects. Either default the slice to 100 or fix this description to match.

src/tools/snapshotManagement.ts
```diff
@@ -0,0 +1,245 @@
...
+  return match ? Number(match[1]) : null;
+}
+
+/**
+ * The next version number for a project: one above the highest ANY plugin
+ * backup in it has ever shown, never a number already used. `floor` carries the
+ * highest version this session issued, so deleting the newest backup and
+ * creating another cannot hand the same number to two different snapshots.
+ */
+export function nextBackupVersion(snapshots: any[], floor = 0): number {
```
dshire, line 154: **[author's choice — probably just document]** The never-reuse guarantee is per-session only: the floor is in-memory, so delete-highest-then-create in a *new* session (or two concurrent sessions listing before either creates) can mint the same `v<N>` twice — names stay unique via the timestamp, but "restore v3" (the reference style the skill teaches) becomes ambiguous. Unavoidable without persistent state, and the docstring already scopes the claim honestly. Suggest one sentence in the SKILL.md: when two backups share a version, disambiguate by timestamp/id rather than version.

dshire (changes requested):
Round 2. First: **every item from the previous review is verified fixed** — I checked the code and the new tests, not just the resolved threads. The gate redesign (pre-call hold + `decline` op) is *better* than what was asked for: the offer now structurally precedes the first mutation, the anti-deadlock trip-once choice is sound and documented, and several fixes went beyond the request (pagination-following list, shared deadline, anti-rot classification test, backward-compatible marker matching). Typecheck clean, 553/553 passing, and the outcome-safety tests that assert "exactly one delete, no create POST" on the halt paths are exactly the right kind of test. Really nice round.

The fresh pass over the new code found a handful of edge issues — **none backup-destroying**. Three are worth fixing before merge (marked **[fix before merge]** inline): the gate's session-global scoping (a decline for project A releases the gate for project B), new *projects* not being exempted the way new agents are, and `created: false` contradicting `outcomeUnknown: true` on the poll-failed path. The rest are **[author's choice]** — take them or leave them with a comment.

After those three, this is mergeable from my side. (A real destructive `restore` on a disposable project before merge, as your test plan already notes, still stands.)

src/tools/handlers.ts
cu4nt0m: Fixed in 3db54f1. The answer is now recorded per project (`snapshotCreatedForProject` / `backupDeclinedForProject`) and the hold is once per project. Rather than resolving with an API call, the gate reads a `projectOfResource` map filled from the reads the model already makes (`list_resources`, `get_resource`) plus what this session created — so `update_ai_agent { aiAgentId }` is scoped correctly whenever the agent has been seen. When the project genuinely cannot be determined the old session-wide fallback still applies, which is the documented residual leak. Tests cover the sandbox-then-production case, including via an agent-only call.

src/tools/handlers.ts
cu4nt0m: Fixed in 3db54f1. The projectId joins `resourcesCreatedThisSession` when `createdProject` is true, and `targetsNewResource` now checks `projectId` — matched only against ids actually recorded, so an existing project passed by a later call cannot false-match its way past the gate.

src/tools/handlers.ts
cu4nt0m: Fixed in 3db54f1. `created` is `null` on the unknown path, and the outcome-safety test pins it. The tool description now says so too, so a caller does not have to infer it.

src/tools/handlers.ts
cu4nt0m: Taken in 3db54f1 — first option. `delete_resource` for `endpoint` and `knowledge_store` is no longer gated: a snapshot cannot bring either back, so the offer was promising a rollback that does not exist, and skipping it matches the `manage_webchat` / `manage_knowledge` exemptions.

src/tools/handlers.ts
cu4nt0m: Taken in 3db54f1. Gated tools are checked against their zod schema before the hold is consumed, so an invalid call falls through to its own validation error and the corrected retry is still protected. Trip-once is unchanged otherwise.

src/tools/snapshotManagement.ts
cu4nt0m: Taken in 3db54f1 — `version: isAutoBackup(raw) ? parseBackupVersion(raw?.name) : null`, with a filter test for a hand-named `[AI Backup] v9 test`.

src/tools/definitions.ts
cu4nt0m: Taken in 3db54f1. The slice defaults to 100 again rather than returning everything, so the description is accurate and an omitted limit cannot flood the result on an installation with a raised cap.

src/tools/snapshotManagement.ts
cu4nt0m: Documented in 3db54f1, as suggested. The skill now says numbers are unique per session but not across sessions, and to identify a backup by timestamp or id if two ever share a version.

cu4nt0m:
Round 2 addressed in `3db54f1`. All three **[fix before merge]** items plus every **[author's choice]** one — taken rather than declined, since each was a real inconsistency.

**Fix before merge**
- **Gate scoping** — the answer is now per project (`snapshotCreatedForProject` / `backupDeclinedForProject`), held once per project. Instead of an API resolve step, the gate consults a `projectOfResource` map filled from reads the model already makes (`list_resources`, `get_resource`) and from what the session created, so `update_ai_agent { aiAgentId }` is scoped whenever that agent has been seen. Where the project truly cannot be determined, the session-wide fallback remains — that residual is documented in the code.
- **New projects** — `create_ai_agent` records the projectId it created, and `targetsNewResource` checks `projectId`, matched only against recorded ids so an existing project cannot false-match past the gate.
- **`created` on the unknown path** — now `null`, never `false`. Pinned by a test and stated in the tool description.

**Author's choice**
- Gated tools are schema-checked before the one-shot hold is consumed, so an invalid call gets its validation error and the corrected retry is still protected.
- `delete_resource` for `endpoint` / `knowledge_store` is no longer gated — a snapshot cannot restore either, so the offer promised a rollback that does not exist.
- `version` is reported only for real plugin backups, so a hand-named `[AI Backup] v9 test` no longer contradicts `isPluginBackup: false`.
- `list` defaults to 100 again, matching the documented default.
- The skill documents that version numbers are unique per session but not across sessions, and says to disambiguate by timestamp or id.

New tests: decline on one project does not release another (including through an agent-only call), hold-once-per-project, no hold for a project created this session, no hold for snapshot-excluded deletions, invalid args do not consume the hold, `created: null` on the unknown path, and the version guard. 559 tests pass, `tsc` clean.

The destructive `restore` on a disposable project is still outstanding on my side — I'll run it before merge and report back here.

src/tools/handlers.ts
```diff
@@ -4812,6 +4999,974 @@ export class ToolHandlers {
...
+      for (const item of items) this.rememberProjectOf(item);
+    }
+    this.rememberProjectOf(result);
+  }
+
+  /**
+   * The project a gated call acts on, or null when it cannot be told. Pure
+   * lookup — the gate never spends an API call to answer this.
+   */
+  private projectForCall(args: any): string | null {
```
dshire, line 5052: **[fix before merge]** The per-project scoping only activates when this lookup can resolve the project — and the map is fed from *filtered* results, where only the **agent** filter exposes `projectId`; the flow and endpoint filters strip `projectReference`, so pre-existing flows/endpoints never enter `projectOfResource`. Net effect: the exact cross-project leak the round-2 fix targeted survives for flow-targeted calls.

Concrete: decline a backup on sandbox project A, then `manage_flow_nodes { operation: "update", flowId }` on a **production** flow obtained from `list_resources` (flow items carry no projectId, so nothing was learned) → `projectForCall` returns null → the global fallback sees a decline exists *somewhere* → production's first change runs with no offer.

Two one-liner options: expose `projectId: r.projectReference ?? r.projectId` in the flow (and endpoint) filters like the agent filter already does, or have `learnProjectIds` also record the `projectId` the `list_resources`/`get_resource` call itself was made with for every returned item. The scoping test suite should get a flow-targeted cross-project case either way.

plugin/skills/snapshot-backups/SKILL.md
```diff
@@ -0,0 +1,207 @@
...
+- So for an agent that uses knowledge, a snapshot is **not a complete backup**. Say
+  this in one line when you offer it, rather than letting the user believe otherwise.
+- Restoring is **irreversible** and deletes resources before recreating them.
+- Snapshots cannot be renamed or edited after creation.
+
+## Supported workflow
+
+### Back up before changing an existing agent
+
+The server enforces this once per session, so you do not have to remember it. The
```
dshire: **[fix before merge — doc only]** This (and the decline wording further down: "nothing is held for the rest of the session", decline "releases the backup gate for the rest of the session") still describes the old session-global gate. Since 3db54f1 the gate is per **project**: an answer for project A does not release project B, and a second project can legitimately be held later in the same session. A skill-loading client reading this will explain decline semantics wrongly and may treat a correct second hold as a server malfunction. Please sweep the three spots to per-project wording (matching what instructions.ts and the tool description now say).

dshire (changes requested):
Round 3 — verified all 8 previous items fixed, with the mechanisms as requested (per-project gate scoping with the passive lookup map, session-created projects exempted, `created: null` on the unknown path, schema-check before the gate, excluded types skipping it, version guard, list default, skill caveat). Typecheck clean, 559/559 green, and the new scoping test suite is exactly right.

Two last items close it out — the first is the round-2 must-fix escaping through a side door, the second is a wording sweep. After these, it's an approve from this side. (Your own test-plan note still stands too: one real destructive restore against a disposable project before shipping.)

plugin/skills/snapshot-backups/SKILL.md
cu4nt0m: Confirmed and fixed in 5b24410. All three spots swept to per-project wording matching `instructions.ts`:

- intro: "enforces this once per **project**"
- step 5: "nothing else is held **for that project**"
- `decline`: "releases the backup gate **for the given project**"

I also added the consequence the old text omitted, since that was the part most likely to cause a misread: a later change to a different project **can** be held again in the same session, and that is correct behaviour rather than a malfunction — so the model should ask again for the new project. Plus a parenthetical for the fallback case, where the server cannot tell which project a call targets and holds once for the session as a whole.

Sweeping for the phrase turned up the same stale claim in two more places you could not see from the diff hunk: the `decline` handler's comment, and — more importantly — the model-facing `note` it returns, which said "No backup will be taken **this session**." That is the same misinformation reaching the model directly rather than via the skill, so it now reads "for this project … a different project touched later still gets its own backup offer." Both corrected in the same commit.

559 tests still pass; typecheck and manifest clean.

src/tools/handlers.ts
```diff
@@ -4812,6 +4999,975 @@ export class ToolHandlers {
...
```
dshire, line 5052: **[follow-up, non-blocking — accepted limitation]** Per-project scoping currently activates only for agents (the agent filter exposes `projectId`; the flow/endpoint filters strip it, so `projectOfResource` never learns pre-existing flows/endpoints and those calls use the session-wide fallback the SKILL.md now documents). Fine to ship as-is given the narrow exposure. For a follow-up PR, the one-liner is: expose `projectId: r.projectReference ?? r.projectId` in the flow (and endpoint) filters — or pass the calling `list_resources`/`get_resource` args' `projectId` into `learnProjectIds` — plus one flow-targeted cross-project case in the scoping test suite.

dshire (approved):
Approving. The SKILL.md sweep is thorough — all three spots now describe the per-project gate, the decline note text was updated too, and the "a second hold on a different project is correct behaviour" line preempts exactly the confusion the old wording would have caused. Typecheck clean, 559/559 green.

One conscious acceptance, recorded so the thread doesn't read as fixed-in-code: the flow/endpoint project-scoping gap from the previous round is **accepted as a documented limitation**, not closed — the skill now honestly describes the session-wide fallback, and the exposure is narrow (multi-project sessions only, and only for calls carrying just a flowId/endpointId for resources not created or read this session; single-project sessions behave identically either way). A follow-up suggestion is inline.

Last reminder from your own test plan before shipping: one real destructive `restore` against a disposable project.

src/tools/handlers.ts
cu4nt0m: planned for next PR, merging.

PR #33, opened 2026-08-20
feat(skills): run red-teaming in the main session and deliver the report

dshire:
## What

Two things, both aimed at making a red-team run end with the user actually seeing the report and being asked the decisions the skill was designed around:

1. **Run red-teaming in the main session, not a subagent.** Deletes the `cognigy-red-team` agent so the `agent-red-team` skill runs in the user's interactive session. The skill is built around decision questions — agreeing the scope contract (Phase 0), approving fixes (Phase 4), and offering the report delivery — and a subagent can't ask any of those, so it silently took the conservative defaults. Discoverability is unaffected: the skill's own description triggers on "red-team / jailbreak-test / probe guardrails," and agent/skill counts derive from the filesystem, so nothing else needed updating.

2. **Deliver the report.** Adds a "Delivering the report" step to the end of Phase 3 so a run ends with the report in front of the user, not a scratchpad path they have to chase:
   - **Offer to open `report.md`** and open it on confirmation (`open` / `xdg-open` / `start "" <path>` / `Invoke-Item`; falls back to printing absolute paths in headless CI/cron). It's offered rather than force-opened because the report holds verbatim jailbreak transcripts — an unannounced window can steal focus or surface that content mid screen-share.
   - **Then offer, once, to publish the report as a shareable Artifact.** Opt-in. Where the client supports it (e.g. Claude Code), `report.md` is published as the Artifact directly — the client themes and wraps the Markdown — so there's no hand-authored HTML. HTML re-render is only a fallback for clients that render HTML but not Markdown.

## Why

Came out of a live red-team session: the report landed in `/private/tmp/.../scratchpad/` and the user had to be told the path and type `! code …` to see it. Separately, testing showed Claude Code often delegates to a subagent, and the red-team subagent couldn't ask the scope/fix/delivery questions the skill depends on — so it just created backups and applied defaults without confirmation. Both problems trace to delivery and decisions belonging to an interactive orchestrator, so the skill now runs there.

## Scope notes

- Skill + agent-removal only — no code, no manifest/count changes (counts derive from the filesystem).
- Delivery is the run's final step: with propose-and-apply, Phase 4 runs first and its re-verify results are folded into `report.md` before delivery, and Phase 6 cleanup must be verified first — the delivered report is always the final, post-remediation one.
- Phrased client-agnostically so non-Artifact clients degrade gracefully.

## Follow-up

The per-OS opener commands in the skill prose duplicate the tested opener in `src/tools/handlers.ts`; deduping into a shared helper is tracked in #36 to keep this PR docs-only.


plugin/skills/agent-red-team/SKILL.md
```diff
@@ -217,6 +217,21 @@ or from a platform `safetySettings` guard. This is what tells the user where a f
...
 `instructions` field.
 
+### Delivering the report
+
+Two closing steps, in this order — the report is not delivered until the user can actually see it:
+
+1. **Open `report.md` in the user's default viewer — every run, without asking** (`open` on macOS,
+   `xdg-open` on Linux, `start` on Windows). Never make the user hunt for the path or type a command to
+   read the result. If the environment is headless (CI/cron, no opener available), print the absolute
+   paths instead.
```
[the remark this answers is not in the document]
dshire: Fixed in 5393efa — the Windows example is now `start "" <path>` in cmd / `Invoke-Item <path>` in PowerShell.

plugin/skills/agent-red-team/SKILL.md
```diff
@@ -217,6 +217,21 @@ or from a platform `safetySettings` guard. This is what tells the user where a f
 produced a distinct platform-guard refusal, say so — it means the posture rests entirely on the
 `instructions` field.
 
+### Delivering the report
+
+Two closing steps, in this order — the report is not delivered until the user can actually see it:
+
+1. **Open `report.md` in the user's default viewer — every run, without asking** (`open` on macOS,
```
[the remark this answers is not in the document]
dshire: Good catch — fixed in 5393efa. Delivery is now the run's final step: with propose-and-apply, Phase 4 runs first and its re-verify results are folded into report.md before it's opened/published, so the delivered report is the final (post-remediation) one, not the stale pre-fix version.

plugin/skills/agent-red-team/SKILL.md
```diff
@@ -217,6 +217,24 @@ or from a platform `safetySettings` guard. This is what tells the user where a f
...
 `instructions` field.
 
+### Delivering the report
+
+Delivery is the **final step of the entire run** — the report is not delivered until the user can
+actually see it. If the scope contract includes fixes, run Phase 4 first and fold the re-verify results
+into `report.md` (and `findings.json`), then deliver that final version; never open or publish the
+pre-remediation report and leave it stale. Two steps, in order:
+
+1. **Open `report.md` in the user's default viewer — every run, without asking** (`open` on macOS,
```
cu4nt0m: **Consent posture:** This is the one unconditional, no-confirmation, machine-touching action in a skill that otherwise gates every consequential action on explicit consent (Phase 0 "state it back to the user, and get confirmation", Phase 4 "Get approval for each change"). Force-opening a report full of jailbreak-probe transcripts can steal focus and expose that content on screen (e.g. mid screen-share). The headless branch below already proves the cheap alternative: print the path — or ask once — as the universal default.

**Feasibility:** "Without asking" also doesn't hold in practice. In Claude Code an un-allowlisted `open`/`start` triggers exactly the Bash permission prompt this line disclaims; in a no-display remote sandbox the opener can "succeed" while opening nothing; and a client whose skill runner has no shell tool can't execute the step at all. Only one exemption (headless CI) is named, with no way to detect it.

plugin/skills/agent-red-team/SKILL.md
```diff
@@ -217,6 +217,24 @@ or from a platform `safetySettings` guard. This is what tells the user where a f
...
+   `xdg-open` on Linux, `start "" <path>` in `cmd` or `Invoke-Item <path>` in PowerShell on Windows).
```
cu4nt0m: **Duplication:** This per-OS opener dispatch — including the non-obvious `start "" <path>` empty-title quoting fix — already exists as tested, try/caught code in `src/tools/handlers.ts:3769-3804` (the render handler). Restating it as prose creates a second copy that will drift: a fix to the handler never reaches this text. Consider extracting the opener into a shared helper / small tool operation and having the skill call that, per the repo's own "extend an existing tool" convention.

cu4nt0m (commented):
Automated code review (adversarially verified — each finding below survived an independent refutation pass; two other candidates were refuted and dropped). 5 confirmed findings and 1 plausible one, all in the delivery step. Inline comments follow.

**Summary**
1. The unconditional "open without asking" step breaks the skill's own consent posture and is infeasible as stated in several shipped clients.
2. The agent file drops the skill's Artifact capability gate.
3. The prescribed hand-authored HTML re-render is unnecessary in Claude Code, where the Artifact tool accepts `report.md` directly.
4. The per-OS opener commands duplicate opener logic that already exists as tested code in `src/tools/handlers.ts`.
5. (Plausible) The delivery section never states that mutation cleanup must precede delivery.

plugin/skills/agent-red-team/SKILL.md
```diff
@@ -217,6 +217,24 @@ or from a platform `safetySettings` guard. This is what tells the user where a f
...
+   Never make the user hunt for the path or type a command to read the result. If the environment is
+   headless (CI/cron, no opener available), print the absolute paths instead.
+2. **Then offer, once, to also publish the report as a shareable Artifact** — a self-contained HTML page
+   the user opens with a single click, no paths or terminal. This is opt-in: ask, don't assume, and only
+   after step 1. Where the client exposes an Artifact / HTML publish (e.g. Claude Code), render `report.md`
```
cu4nt0m: **Efficiency:** In Claude Code — the named example — the Artifact tool accepts `report.md` directly and themes/wraps it automatically, so prescribing a hand-authored, self-contained HTML re-render forces a full re-transcription of a report that includes wide tables and verbatim transcripts, in exactly the primary client. The inlined design rules are also an undocumented subset of the client's actual artifact contract (no `artifact-design` load, no favicon, no md-direct path) and will drift. Suggest: "publish `report.md` as an Artifact where the client supports it; hand-render HTML only where it doesn't."

plugin/skills/agent-red-team/SKILL.md
```diff
@@ -217,6 +217,24 @@ or from a platform `safetySettings` guard. This is what tells the user where a f
 produced a distinct platform-guard refusal, say so — it means the posture rests entirely on the
 `instructions` field.
 
+### Delivering the report
+
+Delivery is the **final step of the entire run** — the report is not delivered until the user can
```
cu4nt0m, line 222: **Plausible gap:** This section presents itself as the run's complete final sequencing ("run Phase 4 first... then deliver") but never cross-references the Mutation protocol's guaranteed-cleanup obligation. A standalone-skill consumer (no agent file, e.g. Codex) could read this as the exhaustive prerequisite list and deliver while a scratch mutation is still un-restored. Largely mitigated by the report template's "Environment changes and restoration" section, but one clause here ("after cleanup has been verified") would close it.

plugin/agents/cognigy-red-team.md
```diff
@@ -15,6 +15,7 @@ You have the Cognigy MCP tools available (`get_resource`, `list_resources`, `man
 4. **Report.** Write `findings.json` and `report.md` from the skill's templates, to the user's scratchpad or a path they name — never into a repository unless asked. Include the What held and Coverage limits sections and the layer attribution.
 5. **Fix and re-verify** — only if the contract permits. Propose exact before/after config text, apply on approval, re-run the failing probes plus a sample of previously passing ones, and report what closed.
 6. **Clean up and prove it.** Restore any mutated config, read it back to confirm it matches, delete scratch resources, and run functional probes proving the agent works and the test artifact is gone.
+7. **Deliver.** As your final action, open the finished `report.md` in the user's default viewer (see the skill's Delivering the report step). The optional Artifact/HTML publish is interactive, so it is the caller's to make, not yours: end your return by giving the report path, noting it's open, and flagging that the caller can offer to publish it as a shareable Artifact.
```
cu4nt0m: **Dropped capability gate:** "flagging that the caller can offer to publish it as a shareable Artifact" restates the skill's step 2 without its conditional — SKILL.md gates the offer on the client exposing an Artifact/HTML publish and says to skip it otherwise. This agent file ships byte-identical to Antigravity/Gemini CLI (`stagePluginDir()` does a plain `cpSync` of `plugin/agents/*.md`), where no Artifact capability exists, so the subagent would tell the caller to offer a publish the client can't perform. Mirror the gate, e.g. "...can offer to publish it as a shareable Artifact where the client supports that."

plugin/skills/agent-red-team/SKILL.md
dshire: Fixed in e77ac47. Delivery now states the path and *offers* to open it, opening on confirmation — no unconditional force-open. Also, the skill no longer runs as a subagent (agent file deleted this PR): it runs in the interactive main session, so it can actually ask.

plugin/skills/agent-red-team/SKILL.md
dshire: Fixed in 3917dbf. Where the client supports it (e.g. Claude Code), `report.md` is now published as the Artifact directly — the client themes/wraps the Markdown — so no hand-authored HTML re-render. HTML is only a fallback for clients that render HTML but not Markdown.

plugin/skills/agent-red-team/SKILL.md
dshire: Fixed in 3917dbf. The delivery step now requires Phase 6 cleanup to be verified before delivering, and never delivers while a scratch mutation is un-restored.

plugin/agents/cognigy-red-team.md
dshire: Moot — the `cognigy-red-team` agent file was deleted in 3917dbf. Red-teaming now runs as a skill in the main session (which can ask the interactive questions a subagent could not), so there is no byte-identical agent file shipped to Antigravity/Gemini to carry a wrong capability claim.

plugin/skills/agent-red-team/SKILL.md
dshire: Agreed, valid DRY point. Deferring to keep this PR docs-only — tracked in #36. A generic "open a local file" op does not map onto the 16 Cognigy-API tools, so a shared internal helper reused by the render handler is likely the better shape than a new tool surface; noted in the issue.

PhotonicDev (approved):
LGTM

PR #38, opened 2026-08-28
feat: attribute plugin changes in Cognigy audit events

cu4nt0m:
## Summary

- Add actor attribution to every write this plugin makes: requests now carry the `X-Ask-AI-Context` header, so Cognigy records `performedBy.actor = "mcp-plugin"` on the resulting audit event and Admin Center → Audit Events distinguishes plugin-made changes from ones people made by hand. Needs Cognigy.AI 2026.17.0 (Cognigy/cognigy#13652), which introduced the field and already ships `mcp-plugin` as a first-class actor value; older versions ignore the unknown header and record the action as before.
- Add an organisation-scoped `audit_event` type to `list_resources` / `get_resource` over `GET /v2.0/auditevents`, with `actor` / `eventType` / `user` filters, so "what did you change?" and "who changed this?" are answerable in-session.
- No breaking changes and no migration. `COGNIGY_DISABLE_AUDIT_ATTRIBUTION=1` opts out of attribution; requests made outside a tool call are byte-identical to before.

Verified end to end against a live 2026.17.0 tenant: a create and a delete issued inside one tool-call scope both came back with `performedBy: { actor: "mcp-plugin", taskId, sessionId }`, sharing one `taskId`, and `list_resources { resourceType: "audit_event", actor: ["mcp-plugin"] }` returned them. Attribution is self-declared provenance for auditability, not a security control — the platform trusts this header from any authenticated caller, which the upstream PR documents as a tracked gap.

## Changes

| File / Component                    | Description                                                                                                                                                                                                                   |
| ----------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `src/utils/actorContext.ts`         | New. Builds the `X-Ask-AI-Context` value from an `AsyncLocalStorage` task id plus a boot-time session id. Documents the two load-bearing details of the platform contract: the actor field on the wire is `type`, not `actor`, and `taskId` is mandatory and must be a string — get either wrong and service-api silently records the action as human-performed. |
| `src/api/client.ts`                 | Set the header in the existing request interceptor, so every platform request is covered including `uploadFile`. Skipped outside a tool call and when attribution is disabled.                                                  |
| `src/index.ts`                      | Enter one task scope per MCP tool call, so all audit events of a single tool call share a `taskId`. `AsyncLocalStorage` rather than a field on the client, which would let concurrent tool calls interleave their ids.          |
| `src/config.ts`                     | New `auditAttribution` config from `COGNIGY_DISABLE_AUDIT_ATTRIBUTION` (on by default).                                                                                                                                        |
| `src/schemas/tools.ts`              | Add `audit_event` to the `list_resources` / `get_resource` enums plus `actor`, `eventType` and `user` filters. Named `eventType` rather than `type` so it cannot be confused with `resourceType` at the call site.              |
| `src/tools/definitions.ts`          | Mirror the enums and document that `audit_event` is organisation-scoped (no `projectId`) and that a missing `performedBy` means a person performed the action.                                                                  |
| `src/tools/handlers.ts`             | `GET /v2.0/auditevents` case and `get_resource` path. A rejected array filter returns a hint naming 2026.17.0 rather than a bare 400; a 403 explains that audit events need an API key with Admin Center access.               |
| `src/tools/filters.ts`              | `audit_event` response filter. Reads the modification chain from `chain`, which is what the live API returns, while accepting the `modifiedResources` name the REST docs use.                                                   |
| `src/instructions.ts`               | One line pointing at the `audit_event` read surface for "what did you change?" questions.                                                                                                                                     |
| `README.md`, `.claude/CLAUDE.md`    | Document the new env var, the user-facing behaviour, and the cross-repo wire contract with its trust caveat.                                                                                                                   |
| `src/__tests__/*`                   | 14 new tests: header shape, absence outside a tool call, opt-out, upload coverage and concurrent-task isolation (against a real local HTTP server); plus the `audit_event` list/read surface, its filters and its error hints.  |

## Test plan


src/tools/handlers.ts
```diff
@@ -2515,6 +2531,50 @@ export class ToolHandlers {
...
+            return withHints(
+              { error: error.message },
+              {
+                likely_cause:
+                  "This Cognigy version predates 2026.17.0, which introduced the repeatable actor[]/type[] filters on GET /v2.0/auditevents.",
+                action:
+                  "Retry list_resources { resourceType: 'audit_event' } without actor / eventType and filter the returned items yourself.",
+              },
+            );
+          }
```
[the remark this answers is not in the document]
cu4nt0m: Fixed in 9357eb8. The hint now only fires when the 400 actually names the `actor` or `type` filter; every other 400 is rethrown, with a test for a rejected `sort` field.

src/tools/filters.ts
```diff
@@ -133,6 +133,41 @@ export const RESOURCE_FILTERS: Record<string, (raw: any) => any> = {
...
+            ...(r.performedBy.sessionId
+              ? { sessionId: r.performedBy.sessionId }
+              : {}),
+          },
+        }
+      : {}),
+    // The live API returns the modification chain as `chain`, even though the
+    // REST docs name the field `modifiedResources` — accept either.
+    ...(() => {
+      const chain = Array.isArray(r.chain) ? r.chain : r.modifiedResources;
```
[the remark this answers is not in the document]
cu4nt0m: Fixed in 9357eb8. `chain` now only wins when it is non-empty, so an empty `chain: []` falls back to `modifiedResources`. Test added.

src/api/client.ts
```diff
@@ -52,6 +64,16 @@ export class CognigyApiClient {
...
+        // Attribute this write to the plugin in Cognigy's audit events. Set
+        // here rather than per call site so every platform request is covered,
+        // including uploadFile's own headers object. Absent outside a tool
+        // call, which leaves the request byte-identical to before.
+        if (this.auditAttribution) {
+          const actorContext = getActorContextHeader();
+          if (actorContext) {
+            reqConfig.headers[ACTOR_CONTEXT_HEADER] = actorContext;
+          }
+        }
```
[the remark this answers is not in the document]
cu4nt0m: Not changing this. The header is inert on reads: service-api parses it into `request.actor`, which is consumed only by the audit middleware when an action is audited, and `createAuditEventObject` writes `performedBy` only for a non-human actor. Restricting by HTTP method would also be wrong in the other direction — several audited platform actions are POSTs that read rather than mutate, and if the platform ever audits a read we want it attributed rather than silently recorded as human-performed. Keeping one insertion point in the interceptor is also what guarantees `uploadFile` is covered.

src/utils/actorContext.ts
```diff
@@ -0,0 +1,85 @@
...
+
+interface TaskContext {
+  taskId: string;
+}
+
+const taskStore = new AsyncLocalStorage<TaskContext>();
+
+/**
+ * Run `fn` under a fresh task id. Scoped per MCP tool call so that every API
+ * request a single tool call fans out to shares one `taskId` — the correlation
```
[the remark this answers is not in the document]
cu4nt0m: The existing wording is correct: the subject is "every API request a single tool call fans out to", which is singular, so "shares" agrees.

dshire (commented):
Automated review (Claude Code) — 6 findings, none blocking. Two candidates were also investigated and refuted: the actor/eventType/user params being accepted for every resourceType matches the schema's existing convention (channel/useCase behave identically and the descriptions say "audit_event only"), and axios's `actor[]=` bracket serialization is exactly the form the platform's OpenAPI spec documents for `GET /v2.0/auditevents`.

src/tools/handlers.ts
```diff
@@ -2515,6 +2531,57 @@ export class ToolHandlers {
...
+              ...(user ? { user } : {}),
+            },
+          });
+          items = res.items ?? res;
+          total = res.total;
+        } catch (error: any) {
+          // Only a 400 that names one of the two filters gets the version
+          // hint. Any other 400 (a rejected sort field, a bad cursor) is a
+          // real error and must not be dressed up as a version problem.
+          const rejectsFilter =
```
dshire: **correctness** — The version-hint gate regex-sniffs a free-text error message with `/\b(actor|type)\b/i`, which misfires in both directions.

*False positive:* on a current 2026.17.0+ platform, `list_resources { resourceType: 'audit_event', actor: ['mcp-plugin'], sort: 'nope:desc' }` with a server detail like `"Field 'sort' must be of type string"` contains the standalone word `type`, so the real error is swallowed and replaced by the misleading "predates 2026.17.0" hint. The existing test *"rethrows a 400 that is unrelated to the filters"* passes only because its hand-picked message avoids the word "type".

*False negative:* `formatError` falls back to `data.detail || data.title` or the raw axios `Request failed with status code 400` — if an old platform's 400 body doesn't literally name `actor`/`type`, the hint written for exactly that case never fires.

A sturdier shape: on any 400 with `actor`/`eventType` present, retry the request without the filters and filter the items client-side (which is exactly what the hint text tells the LLM to do anyway) — then detection accuracy only affects wording, not behavior.

src/index.ts
```diff
@@ -62,31 +63,40 @@ async function main() {
...
-        };
-      } catch (error: any) {
-        logger.error("Tool execution error", {
-          tool: name,
-          error: error.message,
+      // One task id per tool call, shared by every Cognigy API request the
+      // call fans out to, so the audit events of a single tool call group
+      // together. Logged so a local log line can be matched to an audit event.
+      return runWithTask(async () => {
+        logger.info(`Tool call received: ${name}`, {
```
dshire, line 74: **correctness** — Moving `Tool call received` inside `runWithTask` means it now runs only after the rate limiter passes; the `Rate limit exceeded` early return above logs nothing. Before this PR every incoming call logged the tool name, so a throttling incident was visible in server logs — now the rejected calls leave no trace and an operator can't see which tool was being hammered. Suggest a `logger.warn` (with the tool name) in the rate-limit branch.

src/tools/definitions.ts
```diff
@@ -332,6 +333,37 @@ export const tools: ToolDefinition[] = [
           description:
             "Server-side sort as 'field:direction', e.g. 'lastChanged:desc' or 'name:asc'. Sort on any field the resource returns. Not supported for 'tool' (read from the flow chart, not a list endpoint).",
         },
+        actor: {
```
dshire, line 342: **correctness (contract mismatch)** — The advertised JSON schema allows `actor: []` / `eventType: []` (no `minItems`), but the Zod side uses `.nonempty()`, so an empty array — a plausible "no filter" placeholder from a client model — throws a raw ZodError. Either add `minItems: 1` here or treat `[]` as absent in the handler.

src/tools/definitions.ts
```diff
@@ -332,6 +333,37 @@ export const tools: ToolDefinition[] = [
...
+          type: "array",
+          items: {
+            type: "string",
+            enum: ["human", "ask-ai", "mcp-plugin", "system"],
```
dshire: **reuse** — `AUDIT_ACTORS` / `AUDIT_EVENT_TYPES` are exported from `schemas/tools.ts` as the source of truth, but these enum lists re-hardcode the same strings by hand. When the platform grows a new actor value, one list gets updated and not the other: either the LLM is never told the value exists, or it's advertised but rejected by Zod. Import the constants (`enum: [...AUDIT_ACTORS]`) or add a test asserting the lists match.

src/tools/handlers.ts
```diff
@@ -2570,6 +2637,7 @@ export class ToolHandlers {
       // API key belongs to, so `createdBy` / `lastChangedBy` ids can be
       // attributed instead of guessed.
       user: `/v2.0/users/${id}`,
+      audit_event: `/v2.0/auditevents/${id}`,
```
dshire, line 2668: **consistency** — The Admin-Center 403 hint exists only on the list path; `get_resource { resourceType: 'audit_event' }` rethrows a bare `Forbidden` for the identical failure. Consider a small shared helper (or hint mapper) used by both list and get so the org-scoped-permission knowledge lives in one place.

src/tools/filters.ts
```diff
@@ -133,6 +133,46 @@ export const RESOURCE_FILTERS: Record<string, (raw: any) => any> = {
   // this filter did: both only matter for downloading, which this plugin
   // deliberately does not do.)
   snapshot: summarizeSnapshot,
+  // Audit events are verbose and mostly opaque ids. `performedBy` is the point
+  // of this view — it is absent on human-performed events (the platform only
+  // stores it for non-human actors), so a missing key means "a person did
+  // this", and `mcp-plugin` means this plugin did.
+  audit_event: (r) => ({
```
dshire: **simplification** — This filter nests an IIFE inside a spread inside the object literal just for the `chain`/`modifiedResources` fallback, making it the hardest-to-parse entry in `RESOURCE_FILTERS`. A function body does the same job flat:

```ts
audit_event: (r) => {
  const chain = Array.isArray(r.chain) && r.chain.length ? r.chain : r.modifiedResources;
  return { id: rid(r), /* … */ };
},
```

src/tools/handlers.ts
cu4nt0m: Fixed in dc988eb. Took the sturdier shape you suggested: detection is now behavioural rather than textual. A 400 raised while `actor` or `eventType` is set triggers a refetch without those params, and the filters are applied in-process — which is what the hint only *told* the caller to do anyway — so detection accuracy now affects the wording, not the result. If the unfiltered refetch also fails, the 400 was about something else and the original error is rethrown untouched, which covers the `sort: "nope:desc"` false positive without depending on the message text. The response carries a hint saying the filtering happened here and that `total` therefore counts matches within the fetched page only. Tests: client-side fallback for `actor` and for `eventType`, the no-`performedBy`-means-human case, the original 400 surviving a failed retry, and a 400 with no array filter never retrying at all.

src/index.ts
cu4nt0m: Fixed in dc988eb. `logger.warn(\`Tool call rate limited: ${name}\`)` in the rate-limit branch, so a throttled call still names the tool in the server log.

src/tools/definitions.ts
cu4nt0m: Fixed in dc988eb — took the `minItems: 1` option, so the advertised schema and the Zod `.nonempty()` now agree rather than the JSON schema inviting a value Zod rejects.

src/tools/definitions.ts
cu4nt0m: Fixed in dc988eb. `definitions.ts` now imports `AUDIT_ACTORS` / `AUDIT_EVENT_TYPES` from `schemas/tools.ts` and spreads them into the `enum` lists, so there is one source of truth and nothing to drift.

src/tools/handlers.ts
cu4nt0m: Fixed in dc988eb. Extracted `auditEventForbidden(error)` and used it on both the list and get paths, so the org-scoped-permission knowledge lives in one place. Test added for the `get_resource` 403.

src/tools/filters.ts
cu4nt0m: Fixed in dc988eb. Flattened to a function body with the `chain` / `modifiedResources` fallback resolved in a plain `const` first, as suggested — no more IIFE inside a spread.

dshire (approved):

PR #41, opened 2026-09-01
feat: enforce Cognigy Code Node best practices and standards at write…

jacognigy:
## Summary

Enforces Cognigy Code Node best practices and standards at write time, rejecting non-compliant code before it reaches the platform instead of after.

In plain terms: the plugin was generating Code Nodes that could contain risky or deprecated patterns (missing error handling, deprecated functions, code that tries to do things Code Nodes shouldn't do at all). Those problems only showed up after the fact, once the code was already live. This change adds a checkpoint that catches those problems immediately, at the moment the code is written or updated, and blocks it with a clear explanation of what needs to change.

- Added a new validator (`src/tools/codeNodeValidation.ts`) that checks Code Node content against NA Professional Services standards: only approved API methods are allowed, state methods that have been removed or banned are blocked, and Code Nodes can't attempt things they aren't reliably able to do anyway (HTTP calls, module imports, browser-only APIs). It also flags a few risky-but-not-blocking patterns as warnings.
- Enforced that every Code Node wraps its logic in try/catch, using the standardized error-reporting format. Previously, a Code Node could fail partway through and the flow would just continue on as if nothing went wrong, with no record of what broke. Requiring try/catch means any failure is always caught and reported, so it shows up in the flow instead of failing silently.
- Wired this validation into the main code-creation flow (`src/schemas/tools.ts`), so both standalone Code Nodes and the pre/post-processing code on HTTP tools get checked automatically, with invalid code rejected up front and a clear error message returned.
- Extended the same protection to updates on existing Code Nodes (`src/tools/handlers.ts`). Since an update doesn't know the node's type until it looks it up, the check happens after that lookup but still before anything is saved, and the error response includes a ready-to-use template for the required error-handling format. This path also cleans up a couple of fields the server computes on its own so they don't get sent back incorrectly, and double-checks the node after saving to warn if Cognigy itself reports an error on it.
- Made small supporting updates to `src/tools/definitions.ts` and the flow-nodes Skill documentation so the new enforcement behavior is documented for anyone using the tool.
- Added `.claude/settings.json`.

## Test plan

- Updated 15 existing test fixtures (across `tools.test.ts`, `httpTool.test.ts`, and `updateTool.test.ts`) that were written before the try/catch standard existed. These were brought up to the standard using a shared helper built from the module's own standard error template - no test logic or intent was changed, just the sample code the tests use.
- Ran the full test suite: 476 tests passing, no new failures. One pre-existing, environment-dependent failure in `serverLifecycle.test.ts` reproduces the same way on unmodified `main` and is unrelated to this change.
- Type-checked cleanly (`tsc --noEmit`).
- Formatting check passed (`prettier --check`) on all changed files.
- Lint check showed no new errors (33 pre-existing, unrelated parser-config errors on test files carried over from before this change).

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,12 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(npx eslint *)",
+      "Bash(npx tsc *)",
+      "Bash(./node_modules/.bin/eslint --version)",
+      "Bash(./node_modules/.bin/eslint src/tools/handlers.ts)",
+      "Bash(npm install *)",
+      "Bash(npm run *)"
+    ]
+  }
+}
--- a/plugin/skills/flow-nodes/SKILL.md
+++ b/plugin/skills/flow-nodes/SKILL.md
@@ -207,12 +207,55 @@ Run custom **TypeScript** (a single source string — not multiple files, not HT
 
 **Runtime objects available inside the code:**
 
-| Object    | What it is                                                             |
-| --------- | ---------------------------------------------------------------------- |
-| `input`   | The current input — read/write. `input.text`, `input.data`, etc.       |
-| `context` | Session-persistent store. Read/write values that survive across turns. |
-| `profile` | The contact profile.                                                   |
-| `actions` | Helper actions, e.g. `actions.output(text, data)`, `actions.log(...)`. |
+| Object    | What it is                                                                                   |
+| --------- | --------------------------------------------------------------------------------------------- |
+| `input`   | The current input — read/write. `input.text`, `input.flowName`, `input.sessionId`, etc.       |
+| `context` | Session-persistent store. Read via `context.x`; write and delete only through `api.*` below — never assign to it directly (`context.x = value` does not persist the way Cognigy expects). |
+| `profile` | The contact profile.                                                                           |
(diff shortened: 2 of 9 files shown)


pb-crackers (approved):
Great changes!

dshire:
Thanks for putting this together, and for the thorough write-up. We tested each claim on a live tenant and checked it against the Cognigy docs, and about half of it holds up: `api.httpRequest()`, `fetch()`, and `require()`/`import` really are unavailable in Code Nodes, the state methods are removed, and dot-path `api.deleteContext()` really does silently no-op. Those are valuable and we kept them.

The rest is an internal convention from a services team rather than a platform constraint, and some of it is incorrect: `context.x = value` does persist for the session, `toLocaleString()` and `Intl` work at runtime (the editor underline is the browser type checker), the `api.*` allowlist rejects around twenty documented methods, and the documented modules load as globals rather than through `require`. The mandatory try/catch shape is also a house style, not something the platform needs. Our bar for a gate in this plugin is that the platform itself must fail the code, with a docs citation to back it. Anything the platform supports must not be blocked.

We have opened #45 with the platform-backed subset: the four rejections and two warnings above, each cited and covered by tests, plus the corrected skill documentation. Closing this one in favour of that. If you spot a platform constraint we missed, a small PR with a docs link and a test is very welcome.

PR #42, opened 2026-09-02
feat(tools): support the LLM Prompt node on explicit user request

dshire:
## Summary

- Add support for Cognigy's **LLM Prompt node** (`llmPromptV2`) across the tool surface — creating, reading, updating, and deleting the node, plus full tool support (`llmPromptTool` / `llmPromptMCPTool`) under it, all verified against the live chart API
- The **AI Agent node stays the only default**: every LLM-facing description (tool definitions, always-on instructions, skills) states the LLM Prompt node is used solely when the user explicitly asks for it by name — never offered, never suggested, never a fallback
- Flows driven by an LLM Prompt node have **no agent resource**, so `create_tool` / `update_tool` / `delete_resource` / `list_resources` now accept `flowId` as the addressing alternative to `aiAgentId` (which is unchanged for normal agents)
- Harden LLM auto-assignment for the new mode: prefer a **connected, non-embedding** model — a connectionless LLM fails silently under the node's default `errorHandling: "continue"` with an empty message, which presents as an empty agent response

## Changes

| File / Component                        | Description                                                                                                                                                                                                                                                                       |
| --------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `src/tools/nodeRegistry.ts`             | New `llmPrompt` registry entry (type `llmPromptV2`, `@cognigy/basic-nodes`, requires `config.prompt`) with explicit-request-only steering in its summary                                                                                                                            |
| `src/tools/handlers.ts`                 | `create_ai_agent { agentNodeType: "llmPrompt", systemPrompt }` provisions project + flow + llmPromptV2 node + REST endpoint without an `/v2.0/aiagents` resource; removes the backend's placeholder `unlock_account` tool; tool handlers resolve flows via `aiAgentId` **or** `flowId` and parent tools on `aiAgentJob` first, else `llmPromptV2`; `knowledge`/`send_email` rejected under LLM Prompt (descriptors don't exist there); `llmPromptTool` joins the appendChild→append rewrite set; `manage_flow_nodes create` of an llmPromptV2 cleans up its placeholder tool |
| `src/schemas/tools.ts`                  | `agentNodeType`/`systemPrompt` on create_ai_agent (with a refine rejecting `knowledgeStoreReferenceId` in llmPrompt mode); `aiAgentId`-or-`flowId` refines on create_tool/update_tool; `flowId` on delete_resource/list_resources                                                    |
| `src/tools/definitions.ts`              | LLM-facing contract updates with defensive steering language ("ONLY when the user explicitly asked… never offer, suggest, or fall back")                                                                                                                                            |
| `src/instructions.ts`                   | One always-on hard rule: AI Agent node is always the default; llmPrompt only on explicit request; flowId addressing for those flows                                                                                                                                                 |
| `plugin/skills/flow-nodes/SKILL.md`     | LLM Prompt exception to the "no top-level nodes" rule + full config reference (verified against a live node), incl. the freeform-prompt caveat: persona, job, and guardrails all live in one prompt                                                                                 |
| `plugin/skills/tools-setup/SKILL.md`    | flowId addressing section for tools under an LLM Prompt node                                                                                                                                                                                                                        |
| `plugin/skills/agent-creation/SKILL.md` | Explicit-request-only section for building an agent on an LLM Prompt node                                                                                                                                                                                                           |
| `plugin/skills/troubleshooting/SKILL.md`| Two verified empty-response causes: endpoint-config propagation + per-session caching right after creation, and connectionless LLMs failing silently                                                                                                                                |
| `src/__tests__/llmPromptNode.test.ts`   | New suite (16 tests): llmPrompt provisioning, placeholder cleanup, connected-LLM preference, rollback, flowId-addressed tool CRUD, parent preference when both node types exist, registry create + cleanup                                                                          |

Hard-won details from live verification (Meridian (Test) tenant, all scratch resources deleted): the backend auto-creates a non-deletable `llmPromptDefault` branch plus a placeholder `unlock_account` tool under every new `llmPromptV2`; the resolve descriptors are shared (`aiAgentToolAnswer`, `aiAgentJobCallMCPTool` — `llmPromptCallMCPTool` does not exist); end-to-end flow answered as prompted over the REST endpoint after the connected-LLM fix.

No breaking changes: `aiAgentId` remains required-by-refine when `flowId` is absent, so existing callers are unaffected.

## Test plan


src/tools/handlers.ts
```diff
@@ -1766,6 +1824,225 @@ export class ToolHandlers {
...
+        const llmItems = llmList.items ?? llmList;
+        if (Array.isArray(llmItems) && llmItems.length > 0) {
+          const nonEmbedding = llmItems.filter(
+            (l: any) => !String(l.modelType ?? "").includes("embedding"),
+          );
+          const pool = nonEmbedding.length > 0 ? nonEmbedding : llmItems;
+          const connected = pool.filter((l: any) => l.connectionId);
+          const candidates = connected.length > 0 ? connected : pool;
+          const defaultLlm =
+            candidates.find((l: any) => l.isDefault) ?? candidates[0];
```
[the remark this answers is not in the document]
dshire: Addressed in 718faa9, with one deliberate deviation. Embedding models are now filtered out of the candidate pool entirely, so an embedding-only project gets no pick and reports `llmStatus: "unknown"` with the LLM setup guidance. For connectionless models we still assign the reference but return `llmConnected: false` plus a dedicated warning that explains the empty-reply symptom and how to attach or import a connection. Leaving the node unassigned would make it fall back to the project default, which may be just as broken, and the user would get no pointer to the actual cause. Both paths are covered by tests in `llmPromptNode.test.ts`.

cu4nt0m (commented):
Review of the LLM Prompt node support. Ten findings, most severe first — nine correctness/robustness issues and one structural note. Line-anchored details are in the inline comments; summary:

1. **Embedding exclusion is a `modelType.includes("embedding")` substring test** — misses `amazon.titan-embed-*` / `Pharia-1-Embedding-*`; can wire the agent to an embedding model while reporting `llmStatus: "configured"`. (handlers.ts:558)
2. **`!llmConnected` warning is unreachable when the knowledge tool failed** — the pre-existing early return at 1757 wins; `result.llmConnected` never set. (handlers.ts:1775)
3. **Connected-over-`isDefault` silently overrides the project default LLM** — no `llmRefId` in the result; also changes the existing aiAgent path. (handlers.ts:561)
4. **update_tool LLM-Prompt guard: bypassed when `toolType` is omitted (as its own hint suggests), and its GET is un-caught / unused** — knowledge keys dropped silently; raw failures on ordinary flows. (handlers.ts:4251)
5. **Mixed aiAgentJob + llmPromptV2 flow addressed by `flowId` silently attaches to the aiAgentJob**; response has no `parentNodeId`; definitions text promises a refusal that only fires for same-type duplicates. (handlers.ts:3828)
6. **`removePlaceholderTools` now deletes every tool child of the parent by type** — safe only under an unenforced "just-created parent" precondition; deletions unreported, errors swallowed. (handlers.ts:2081)
7. **aiAgent path still uses `identifyFailedStep`**, which reports `"endpoint"` for job-node/entry-node failures; the llmPrompt path got the explicit `step` fix. (handlers.ts:1790)
8. **`immediateOutput: true` + `storeLocation: "stream"`** is contradictory per the Cognigy docs; `flow-nodes/SKILL.md:337` documents the pairing backwards. (handlers.ts:1964)
9. **Duplicate-toolId detection now requires a parent reference in the node-list excerpt** (low confidence; cheap fallback available). (handlers.ts:3917)
10. **`createLlmPromptAgent` duplicates ~250 lines of the aiAgent orchestration**; drift already visible (propagation hint, failed-step mechanism, LLM lookup `limit`). (handlers.ts:1896)

Checked and not flagged: the multi-aiAgentJob disambiguation (a deliberate improvement over main's "first node wins"), the un-hinted 404 on a bad `flowId` (formatted by the client/index.ts like any other platform error), `talk_to_agent` lacking `flowId` (endpointUrl is discoverable via `list_resources endpoint`), Prettier/versioning/commit conventions (all clean).

Smaller cleanups noticed along the way, no action required: `NOT_BOTH_IDS_MESSAGE` is duplicated verbatim in `resolveToolFlow` (unreachable — every feeding schema already refines it); `buildConfigUrl(endpoint)` exists and could replace the inline URL ternaries; the `createFlowFor` `onCreated` callback and the following `flowId = provisioned.flowId` set the same value twice; the "LLM Prompt only on explicit request" rule is now stated in 9 places (+~3.8k chars to definitions.ts per ListTools); `delete_resource 'tool'` gained one extra GET per deletion just to pre-detect `*Default` branches; and `removePlaceholderTools` still falls back to the 200-node list GET whenever the chart found nothing (now at three call sites).

src/tools/handlers.ts
```diff
@@ -484,15 +499,74 @@ const AI_AGENT_TOOL_TYPES = new Set([
...
+ * `connected` tells the caller whether the pick still had to fall back to a
+ * connectionless model.
+ */
+function pickDefaultLlm(
+  llmList: any,
+): { refId: string; connected: boolean } | undefined {
+  const llmItems = llmList?.items ?? llmList;
+  if (!Array.isArray(llmItems) || llmItems.length === 0) return undefined;
+  const pool = llmItems.filter(
+    (l: any) => !String(l.modelType ?? "").includes("embedding"),
```
cu4nt0m: **Embedding exclusion is a substring heuristic and misses documented embedding models.**

`String(l.modelType ?? "").includes("embedding")` is case-sensitive and only matches the literal word. Models the repo itself lists in `plugin/skills/llm-providers/SKILL.md` — e.g. `amazon.titan-embed-text-v2:0` and `Pharia-1-Embedding-4608` — pass the filter, and with a `connectionId` they win outright over an unconnected default (line 561). Result: `create_ai_agent` (both modes) wires the agent node to a model that cannot generate text and reports `llmStatus: "configured"` with no warning — exactly the silent-empty-reply case this helper's docblock says it prevents.

The `list_resources llm_model` path already uses the API's own discriminator (`/new/v2.0/largelanguagemodels?useCase=…`, with `aiAgent` / `promptNode` documented in definitions.ts). Suggest querying with `useCase` here instead of sniffing `modelType`. The PR tests only cover `text-embedding-*` / `custom-embedding-model`.

src/tools/handlers.ts
```diff
@@ -484,15 +499,74 @@ const AI_AGENT_TOOL_TYPES = new Set([
...
+  );
+  if (pool.length === 0) return undefined;
+  const connected = pool.filter((l: any) => l.connectionId);
```
cu4nt0m, line 728: **Connected-over-`isDefault` silently bypasses the project's designated default LLM — and this changes the existing aiAgent path too.**

Main did `find(isDefault) ?? items[0]`. Now with `[{ X, isDefault: true /* no connectionId yet */ }, { Y, connectionId }]` the Job node is bound to Y, `connected: true`, and the result is returned bare: neither create result includes the chosen `llmRefId`, and the only LLM hints are the `unknown` / `!llmConnected` branches. The user's chosen default is replaced by an unrelated model (e.g. a stale test model) with no indication in the response, and fixing X's connection later does not move the agent back.

Either keep honouring `isDefault` and rely on the new `llmConnected: false` warning, or at minimum surface `llm: { referenceId, isDefault, connected }` in the result when the pick differs from the default. (Note the `tools.test.ts` mock had to gain `connectionId` to keep passing — a behaviour change to the pre-existing path riding along with the LLM Prompt feature.)

src/tools/handlers.ts
```diff
@@ -1717,65 +1772,362 @@ export class ToolHandlers {
         });
       }
 
+      if (!llmConnected) {
```
cu4nt0m: **This warning is unreachable when the knowledge tool failed.**

The pre-existing early `return withHints(result, { warning: "Agent created but knowledge tool failed to provision." … })` at line 1757 runs before this block, and `withHints` does not accumulate. So for `create_ai_agent { knowledgeStoreReferenceId }` in a project whose only chat LLM has no connection: `pickDefaultLlm` → `connected: false`, PATCH succeeds → `llmStatus: "configured"`, knowledge POST throws (swallowed) → line 1757 returns with only the knowledge warning. `result.llmConnected` is never set and the "do not call talk_to_agent" guidance is lost in the one case it matters. `createLlmPromptAgent` has no intervening return, so only this path is affected.

Suggest setting `result.llmConnected` / computing the LLM warning before the knowledge-failure return (or merging hints).

src/tools/handlers.ts
```diff
@@ -1717,65 +1772,362 @@ export class ToolHandlers {
...
+      }
+
       return result;
     } catch (error: any) {
-      const rolledBack: string[] = [];
-      const rollbackFailed: string[] = [];
+      return withHints(
+        {
+          failed: {
+            step: identifyFailedStep(agentId, flowId, endpointId),
```
cu4nt0m: **Failed-step attribution is still wrong on this path; the fix exists in the sibling method.**

`identifyFailedStep(agentId, flowId, endpointId)` returns `"endpoint"` whenever `agentId` and `flowId` are set and `endpointId` is not — i.e. for a Step 4 `POST /chart/nodes` (aiAgentJob) failure or a `retryGetEntryNode` failure, both of which happen before the endpoint is created. The PR's own comment in `createLlmPromptAgent` ("the node is created BEFORE the endpoint here, so the failed step cannot be derived from which ids are set") applies word-for-word here, and that method got an explicit `step` tracker. Pre-existing on main, but this function was refactored in the same hunk — worth using the same `step` variable in both (the `"node"` branch of `identifyFailedStep` is effectively unreachable today).

src/tools/handlers.ts
```diff
@@ -1717,65 +1772,362 @@ export class ToolHandlers {
...
+  }
+
+  /**
+   * create_ai_agent with agentNodeType "llmPrompt": provision project + flow +
+   * LLM Prompt (llmPromptV2) node + REST endpoint. There is NO /v2.0/aiagents
+   * resource in this mode — the "agent" is just the flow, so update_ai_agent
+   * does not apply (the prompt is edited via manage_flow_nodes update) and
+   * tools are addressed via create_tool { flowId }.
+   */
+  private async createLlmPromptAgent(
```
cu4nt0m, line 2128: **~250 lines of the aiAgent orchestration are duplicated rather than parameterised.**

The `GET /largelanguagemodels` + `pickDefaultLlm` try/catch (1636-1659 vs 1934-1949), the `llmStatus` derivation, the `resourcesCreatedThisSession` / `rememberProjectOf` registration, and the `unknown` / `!llmConnected` hint blocks (already worded differently) are copies, and the two paths use incompatible failed-step mechanisms (`identifyFailedStep` vs `step`). Drift is already visible: the endpoint-propagation hint ("retry with a NEW sessionId") exists only in the llmPrompt `usageHint` though it applies to every fresh REST endpoint; the aiAgent path kept the misreporting `identifyFailedStep`; and the other `largelanguagemodels` call at ~5836 passes `limit: 100` while these two pass none.

Since `createProjectFor` / `createFlowFor` / `createRestEndpointFor` / `rollbackProvisioned` were already extracted, consider one `provisionAgentFlow({ createDrivingNode })` pipeline with a mode-parameterised hint block, so future provisioning changes are made once.

src/tools/handlers.ts
```diff
@@ -1717,65 +1772,362 @@ export class ToolHandlers {
...
+        {
+          mode: "append",
+          target: entryNode._id,
+          type: "llmPromptV2",
+          extension: "@cognigy/basic-nodes",
+          label: data.name,
+          config: {
+            prompt: data.systemPrompt?.trim() || data.description?.trim() || "",
+            storeLocation: "stream",
+            immediateOutput: true,
```
cu4nt0m: **`immediateOutput: true` is dead/contradictory config in `stream` mode — and the skill documents it backwards.**

Per the Cognigy LLM Prompt node docs ("Storage & Streaming Options"), *Output result immediately* "appears when you select either Store in Input or Store in Context"; in Stream to Output mode the result already streams and the flag has no meaning. Every llmPrompt-mode create writes a flag the UI never shows for the chosen mode, and a later switch of `storeLocation` to input/context inherits `immediateOutput: true` nobody chose. `plugin/skills/flow-nodes/SKILL.md:337` states the opposite pairing ("Output the result immediately (with storeLocation: \"stream\")"), so the model will keep emitting this combination via `manage_flow_nodes`.

Suggest dropping `immediateOutput` from the stream payload and correcting the SKILL.md row.

src/tools/handlers.ts
```diff
@@ -1717,65 +1772,362 @@ export class ToolHandlers {
...
+    flowId: string,
+    parentNodeId: string,
+    childType: "aiAgentJobTool" | "llmPromptTool",
+    preferredLocaleId?: string,
+  ): Promise<void> {
+    // Without the parent id the child filters below would match unrelated
+    // nodes (e.g. every node lacking parentId) — better to leave the
+    // placeholder in place than to delete real tools.
+    if (!parentNodeId) return;
+    const isPlaceholder = (n: any) =>
```
cu4nt0m: **`n.type === childType` turns "remove the placeholder" into "delete every tool child of this parent".**

Main matched only the `unlock_account` preview/label/toolId markers. The type clause is now load-bearing (the llmPrompt placeholder is labelled `"Tool"` per the test fixture, so markers never match it), which means this shared helper is safe only under the precondition stated in the docblock — "a just-created parent has no other tool children yet" — and nothing in code enforces it: no cap on deletions, deleted ids are not returned, and all errors are swallowed to `logger.warn`. All three current callers do pass a just-created parent, so there is no live bug today; but any future/retry caller with an existing parent (a repair op, a retried partial create, a `manage_flow_nodes` op re-targeting an existing llmPromptV2) would silently wipe real tools.

Cheap hardening: only apply the type match when it yields at most one child, and/or return the deleted ids so callers can surface them.

src/tools/handlers.ts
```diff
@@ -3389,45 +3773,134 @@ export class ToolHandlers {
...
+          {
+            action: `Pick one of: ${parentCandidates.map((n: any) => `${n._id || n.id} (${n.type})`).join(", ") || "none — the flow has no tool parent node"}.`,
+          },
+        );
+      }
+    } else {
+      // The AI Agent Job node wins when a flow has both parent types (see
+      // findToolParentNode); only several nodes of the preferred type are
+      // ambiguous.
+      const preferred = findToolParentNode(allNodes);
```
cu4nt0m, line 4157: **Mixed flows addressed by `flowId` are silently routed to the aiAgentJob, and the response doesn't say which parent was chosen.**

With one `aiAgentJob` + one `llmPromptV2` (reachable now that `manage_flow_nodes` can add a top-level llmPrompt node to an existing flow), `findToolParentNode` returns the aiAgentJob, `sameType.length === 1`, so the "pass parentNodeId" error at 3856 never fires and the tool is POSTed as `aiAgentJobTool` under the AI Agent node even though the caller used `flowId` (documented as the way to target LLM Prompt flows). The success result is `{ toolId, name, toolType }` — no `parentNodeId`/type — so the model cannot notice. `definitions.ts:633` ("Only needed when the flow has more than one such node (create_tool then refuses to guess…)") promises a refusal that only happens for same-type duplicates.

Suggest treating a mixed-type flow as ambiguous when addressed by `flowId`, and returning `parentNodeId` in the create response (plus rewording the definitions text to "more than one node of the preferred type").

src/tools/handlers.ts
```diff
@@ -3436,9 +3909,12 @@ export class ToolHandlers {
         : undefined;
 
     if (requestedToolId) {
-      const duplicateTool = (Array.isArray(allNodes) ? allNodes : []).find(
+      // Scoped to the chosen parent: the same toolId under another tool
+      // parent in the flow is a different tool, not a duplicate.
+      const duplicateTool = allNodes.find(
         (node: any) =>
           MCP_MANAGED_TOOL_TYPES.has(node.type) &&
+          nodeParentId(node) === jobNodeId &&
```
cu4nt0m: **Duplicate-toolId detection now depends on the node-list excerpt carrying a parent reference.** (low confidence)

Main matched flow-wide; the new `nodeParentId(node) === jobNodeId` clause returns `undefined` for a missing/null parent, which never equals `jobNodeId`, so on any projection where tool children come back without `parentId` a second tool with the same toolId is created instead of `reusedExisting: true`. The two pre-existing duplicate tests only kept passing after `parentId: ID.node` was added to their mocks. The repo documents `parentId` as present on `/chart/nodes`, so this is low-probability — flagging because the fallback is cheap: match flow-wide when *no* node in the list carries any parent reference.

src/tools/handlers.ts
```diff
@@ -3761,6 +4245,30 @@ export class ToolHandlers {
     const cfg = data.config;
     const toolType = data.toolType;
 
+    // Mirror create_tool's guard: descriptors that don't exist under an LLM
+    // Prompt node (knowledge, send_email) must not be PATCHed onto its tools —
+    // the platform accepts the unknown keys and the tool silently does nothing.
+    if (cfg && toolType && !(toolType in LLM_PROMPT_TOOL_TYPE_MAP)) {
```
cu4nt0m: **The LLM-Prompt guard is bypassed by the case its own hint recommends, and its GET is un-caught.**

1. The guard runs only when `toolType` is knowledge/send_email. With `toolType` omitted (optional — and the error hint at 4278 literally says "or omit toolType"), control falls to the `toolType === "tool" || … || !toolType` branch at 4296: knowledge/send_email keys (`knowledgeStoreId`, `topK`, `recipient`) are not mapped, so the call reports `updated: true, updatedFields: []` with no warning — the "silently does nothing" this comment says it prevents — and plain-tool keys (`toolId`/`description`/`parameters`) get PATCHed onto `llmPromptMCPTool` nodes. (The `!toolType` branch is pre-existing for `aiAgentJobMCPTool`; the PR extends it to the new family and adds an incomplete guard.)
2. The `GET /chart/nodes/{toolNodeId}` has no try/catch and its result is not reused (the only other reference is the PATCH at 4325). On an ordinary aiAgent flow — where the guard can never fire, since knowledge/send_email tools cannot exist under an LLM Prompt node — every knowledge/send_email update now pays an extra round-trip and a non-retried GET failure throws raw where main would have PATCHed.

Suggest: fetch the node once (or only when `data.flowId` was given, the only way to reach an LLM Prompt flow), catch the GET, and dispatch on `node.type` rather than the caller's `toolType`.

src/schemas/tools.ts
```diff
@@ -8,12 +8,23 @@ const paginationSchema = {
...
+    systemPrompt: z.string().optional(),
+  })
+  .refine(
+    (d) => !(d.agentNodeType === "llmPrompt" && d.knowledgeStoreReferenceId),
+    {
+      message:
+        "knowledgeStoreReferenceId is not supported with agentNodeType 'llmPrompt' — LLM Prompt nodes only support 'tool', 'mcp', and 'http' tools.",
+      path: ["knowledgeStoreReferenceId"],
+    },
+  );
```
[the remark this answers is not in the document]
cu4nt0m: Already enforced at schema-parse time — `createAiAgentSchema` carries a `.refine` that rejects `agentNodeType: "llmPrompt"` unless a trimmed `systemPrompt` or `description` is present (it landed in 6cb5f14, before this comment was resolved). 720caf4 adds the missing coverage in `src/__tests__/schemas.test.ts`: llmPrompt with neither field, llmPrompt with a whitespace-only systemPrompt, llmPrompt with only a description, and aiAgent with neither (still valid).

src/tools/handlers.ts
```diff
@@ -1766,6 +1824,225 @@ export class ToolHandlers {
...
+        `/v2.0/flows/${flowId}/chart/nodes`,
+        {
+          mode: "append",
+          target: entryNode._id,
+          type: "llmPromptV2",
+          extension: "@cognigy/basic-nodes",
+          label: data.name,
+          config: {
+            prompt: data.systemPrompt ?? data.description ?? "",
+            storeLocation: "stream",
```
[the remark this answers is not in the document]
cu4nt0m: Fixed in 720caf4. `createLlmPromptAgent` now takes the first field that is non-blank after trimming (`data.systemPrompt?.trim() || data.description?.trim() || ""`), so a whitespace-only `systemPrompt` falls through to `description` instead of provisioning an effectively empty prompt. Covered by a test, and the schema refinement rejects the both-blank case before any resource is created.

src/tools/handlers.ts
cu4nt0m: Fixed in 720caf4 as suggested. Candidate models now come from the platform discriminator — `GET /new/v2.0/largelanguagemodels?useCase=aiAgent` on the aiAgentJob path and `?useCase=promptNode` on the LLM Prompt path, matching the `list_resources llm_model` branch. If that filtered request fails (older platform), it refetches unfiltered and filters locally, mirroring the audit-event precedent rather than throwing. The local filter is now case-insensitive `/embed/i` over `modelType` and `openAICompatible.customModel`, so `amazon.titan-embed-text-v2:0` and `Pharia-1-Embedding-4608` are excluded; both are covered by tests, including the case where they carry a connection and previously won outright.

src/tools/handlers.ts
cu4nt0m: Fixed in 720caf4, taking the second option. The connected-over-`isDefault` preference stays (it is what prevents the silent empty reply), but both create results now always carry `llm: { referenceId, isDefault, connected }`, and when a text-capable project default was passed over for having no connection the response gets a warning naming the skipped default and the model used instead, plus an action for moving back (attach a connection, then `update_ai_agent { jobConfig: { llmProviderReferenceId } }` / `manage_flow_nodes update`). An embedding default never triggers that hint, since it is not in the pool. Tests cover both paths.

src/tools/handlers.ts
cu4nt0m: Fixed in 720caf4. The LLM branches no longer return; they assign into a local hint set (same precedence as before) and set `result.llmConnected` before the knowledge-failure return, which now returns both hint sets merged through a new `mergeHints` helper — `withHints` still overwrites, so the merge happens at the call site rather than changing behaviour for other callers. `create_ai_agent { knowledgeStoreReferenceId }` against an unconnected LLM now reports both warnings, both actions and the "do not call talk_to_agent" guidance. Test added.

src/tools/handlers.ts
cu4nt0m: Fixed in 720caf4. The aiAgent path now uses the same explicit `step` tracker as `createLlmPromptAgent`, with identical stage names, and reports it as `failed.step`; a chart-node or `retryGetEntryNode` failure is reported as `step: "node"` instead of `"endpoint"`. `identifyFailedStep` had no remaining callers and is removed. Test added for the node-stage failure.

src/tools/handlers.ts
cu4nt0m: Agreed on the analysis, but deliberately not done in this follow-up — consolidating the two paths into one `provisionAgentFlow({ createDrivingNode })` pipeline is a refactor, and doing it in the same change as nine correctness fixes would make both harder to review and to revert. The concrete drift items you list are fixed individually in 720caf4: both paths now share the `useCase` fetch helper, the same explicit `step` tracker (so `identifyFailedStep` is gone), and the same `llm: {...}` reporting plus the skipped-default hint. Tracking the consolidation separately.

src/tools/handlers.ts
cu4nt0m: Fixed in 720caf4 as suggested. `immediateOutput` is dropped from the stream-mode payload, with a comment recording that the UI only offers "Output result immediately" for Store in Input and Store in Context, and `plugin/skills/flow-nodes/SKILL.md:337` now documents the correct pairing (applies with `storeLocation: "input"` or `"context"`; meaningless with `"stream"` and must not be sent). A test asserts the created config carries no `immediateOutput`.

src/tools/handlers.ts
cu4nt0m: Hardened in 720caf4, both halves. The marker match (preview/label/`config.toolId` === `unlock_account`) always deletes; the bare node-type match deletes only when it selects exactly one child, and otherwise deletes nothing by type and logs that the parent already has real tool children, so a repair or retry caller targeting an existing parent cannot wipe real tools. `removePlaceholderTools` now returns the deleted node ids, and `manage_flow_nodes create` surfaces them as `placeholderToolsRemoved` (documented in `definitions.ts`); the `create_ai_agent` callers ignore the return, since their parents are seconds old. The docblock states the enforced cap instead of the previous unenforced precondition. Two tests added.

src/tools/handlers.ts
cu4nt0m: Fixed in 720caf4, all three parts. A flow addressed by `flowId` whose tool-parent candidates span both families is now refused with the candidates listed (`{ nodeId, type, label }`) and a "pass parentNodeId" action, in the same shape as the same-type-duplicate refusal; addressing by `aiAgentId`, single-candidate flows and an explicit `parentNodeId` are unchanged. Create responses (new, reused-existing and HTTP) now report `parentNodeId` and `parentNodeType`, and `definitions.ts:633` is reworded to "more than one node of the preferred type, or when a flow addressed by flowId holds both an AI Agent Job node and an LLM Prompt node". Tests cover the refusal, the reported parent, and the `aiAgentId` path.

src/tools/handlers.ts
cu4nt0m: Fixed in 720caf4 with the cheap fallback you suggested. Duplicate-toolId matching stays parent-scoped when any node in the list carries a parent reference, and falls back to flow-wide matching when none does, with a comment recording why. Two tests pin both halves.

src/tools/handlers.ts
cu4nt0m: Fixed in 720caf4 along the lines you suggested. The family decision now comes from the fetched node's own `type` rather than the caller's optional `toolType`, so knowledge/send_email keys aimed at an LLM Prompt tool node are refused with `updated: false` and an actionable warning even when `toolType` is omitted — and those same keys now do apply when the node really is a `knowledgeTool` or `sendEmailTool`, which the old branch dropped. The node is read at most once, inside a try/catch (a read failure logs at debug and degrades to the previous PATCH behaviour), that response is reused for the PATCH, and the read only happens when the request carries a key that exists in one family and not the other, so ordinary config updates pay no extra round-trip. Also, `updated: true, updatedFields: []` can no longer be returned: an update that applies nothing reports `updated: false` with a warning. Tests added in `updateTool.test.ts` and `llmPromptNode.test.ts`.

plugin/skills/flow-nodes/SKILL.md
```diff
@@ -11,6 +11,8 @@ Use `manage_flow_nodes` to add logic nodes **inside tool branches only**. Nodes
 
 **Voice exception — Set Session Config:** The one node that _should_ run before the AI Agent node is a `setSessionConfig` (Set Session Config) node, and only in **voice** flows. It applies per-session speech settings (barge-in, ASR, STT/TTS, input timeouts) and must be the **first** node. The `audit_voice_agent` tool checks for this and can create it by `prepend`ing before the AI Agent node. Do not add any other pre-agent nodes.
 
+**LLM Prompt exception — explicit request only:** The `llmPrompt` node (`llmPromptV2` — a raw LLM call driven by a freeform system prompt) is the one node type that legitimately lives at the **top level** of a flow, in flows that have **no AI Agent node at all**. It is supported ONLY when the user **explicitly asks for an LLM Prompt node by name**. NEVER offer it, never use it as a fallback for the AI Agent node, and never ask the user to choose between the two — the AI Agent node (via `create_ai_agent`) is always the default for anything agent-shaped. Reading and updating llmPromptV2 nodes that already exist in a flow is always fine. See the [LLM Prompt section](#llmprompt--llm-prompt-explicit-request-only) for config.
```
PhotonicDev, line 14: Man the prompts are really verbose can we just at least get rid of redundant info and filler words, it's really not gonna affect that much if we just make it less verbose

PhotonicDev (dismissed):
LGTM, just one comment

dshire:
I agree with Nerijus, let's reduce verbosity before merge.

cu4nt0m (approved):

PR #47, opened 2026-09-07
fix(api): tunnel Cognigy requests through corporate proxies

cu4nt0m:
## Summary

- Fixes the reported bug where, behind a corporate proxy, the tool list loads but **every** tool call returns `{"error":"API request failed","status":500}` — and setting `HTTP_PROXY`/`HTTPS_PROXY` changes nothing.
- Root cause is in axios, not the customer's network: axios reads the proxy env vars but its Node adapter never opens a `CONNECT` tunnel, so an `https://` Cognigy call leaves the process as a plaintext HTTP request asking the proxy to fetch an HTTPS URL. Corporate proxies reject that form and answer with an HTML error page — the platform is never reached.
- Adds real tunnelling agents (`proxy: false` + `http(s)-proxy-agent`) for every outbound request, and makes the error message name what actually responded instead of collapsing to a generic string.

## Root cause

`node_modules/axios/lib/adapters/http.js` → `setProxy()`:

```js
options.hostname = proxyHost;   // send to the proxy
options.path = location;        // absolute-form URI
options.protocol = proxy.protocol; // http: — so no client-side TLS at all
```

The request the proxy expects is `CONNECT api-*.cognigy.ai:443`, followed by an end-to-end TLS handshake through the resulting tunnel. What it received instead was `GET https://api-*.cognigy.ai/v2.0/... HTTP/1.1` in cleartext. Squid, Zscaler, BlueCoat and similar refuse that and return a 5xx HTML page.

`formatError` then looked for Cognigy's `detail`/`title` fields, found neither, and fell back to `"API request failed"` — which is why the report carried no clue that a proxy was involved.

The tool list works because `ListTools` makes no HTTP request, which is the tell for this class of failure.

## Changes

| `src/utils/proxy.ts` | New. Resolves the proxy for a target URL via `proxy-from-env` (axios' own dependency, so upper/lower-case vars, `ALL_PROXY` and `NO_PROXY` behave as users already expect) and returns axios options with `proxy: false` plus pooled `HttpProxyAgent`/`HttpsProxyAgent`. Rejects unusable values (no host, SOCKS) with a warning and a direct connection rather than tunnelling to a nonsense host. Redacts credentials before logging. |
| `src/api/client.ts` | Applies those options to the axios instance, so every platform request — uploads included — is tunnelled. `describeUnexpectedBody` replaces the bare `"API request failed"` with the HTTP status, a whitespace-collapsed 300-char snippet of the body, and, when a proxy applies to that request, the redacted proxy URL plus the `NODE_EXTRA_CA_CERTS` hint. |
| `src/tools/handlers.ts` | `talk_to_agent`'s standalone `axios.post` gets the same wiring, resolved per call — the endpoint host differs from the API host and may be excluded by `NO_PROXY` independently. |
| `src/index.ts`, `src/utils/proxy.ts` | Logs the effective proxy once at boot, so support can tell from the client's MCP log whether the engine picked the settings up. GUI clients (Desktop, Antigravity) start the engine with a minimal environment and often do not inherit shell proxy variables. |
| `src/types/proxy-from-env.d.ts` | Four-line declaration for the one untyped function used, instead of a dev dependency. |
| `package.json` | Adds `https-proxy-agent`, `http-proxy-agent` (pinned `9.1.0`) and `proxy-from-env` (`1.1.0`, already in the tree via axios). |
| `README.md`, `plugin/skills/troubleshooting/SKILL.md`, `.claude/CLAUDE.md` | Documents the env vars, the TLS-inspection case, and the "tool list works but every call fails" symptom. |

## Test plan

Automated:

- `npm test` — 659 passed, 3 skipped. 23 new tests across `src/__tests__/proxy.test.ts` (agents attached, `proxy: false` always, lower-case vars, `NO_PROXY` honoured, credentials passed through, agents pooled, SOCKS and hostless URLs refused, credentials redacted) and `src/__tests__/apiErrorMessages.test.ts` (status + snippet, HTML/JSON/Buffer bodies, truncation, proxy named only when one applies, no throw on a config with no usable URL).
- `npm run build`
- `npm run check:manifest`

Manual — end-to-end against a local proxy that accepts **only** `CONNECT` and answers absolute-form requests with a 501 HTML page, fronting an HTTPS target with a self-signed cert trusted via `NODE_EXTRA_CA_CERTS`:

```
OLD: status=501 body=<html><head><title>ERROR</title></head><body> <p>ERROR: The requested URL could…
NEW: {"items":[{"name":"tunnelled ok","url":"/v2.0/projects"}]}

proxy: refused absolute-form request -> https://localhost:8443/v2.0/projects
proxy: CONNECT -> localhost:8443
```

The first two lines are the same env vars through the old and new code paths; the proxy log shows the request shape changing from rejected absolute-form to an accepted tunnel. This also exercises the `NODE_EXTRA_CA_CERTS` path, since the tunnelled TLS handshake verifies against that CA.

- None. With no proxy configured the options are `{ proxy: false }` and no agents, which matches what axios does anyway when no proxy variable is set — the direct-connection path is unchanged.

## Support answer for the reported issue

Running behind a corporate proxy was **not** supported and the failure is a bug on our side, not customer configuration. After this ships, customers set the proxy in the MCP server `env` block:

```json
"env": {
  "HTTPS_PROXY": "http://proxy.corp.example:8080",
  "NODE_EXTRA_CA_CERTS": "/etc/ssl/certs/corporate-root-ca.pem"
}
```

`NODE_EXTRA_CA_CERTS` is only needed when the proxy inspects TLS, and requires a client restart because Node reads it at startup. Certificate verification is never disabled. HTTP and HTTPS proxies are supported; SOCKS proxies are not, and are now refused with an explicit message rather than failing obscurely.

## Marketplace checklist

- [x] Tool count in `README.md` updated if the tool surface changed — unchanged, still 17 tools
- [x] `README.md` updated if user-facing behavior changed — new "Corporate proxies" section and three env-var rows
- [x] Tool annotations/descriptions reviewed if tool behavior changed — no tool contract changed
- [x] `LICENSE` / marketplace metadata reviewed if submission requirements are affected — unaffected

### DIFF
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -7,7 +7,11 @@ import axios, {
 } from "axios";
 import FormData from "form-data";
 import { logger } from "../utils/logger.js";
-import { getProxyAxiosOptions, redactProxyUrl } from "../utils/proxy.js";
+import {
+  applyProxyToRedirect,
+  getProxyAxiosOptions,
+  redactProxyUrl,
+} from "../utils/proxy.js";
 import { getProxyForUrl } from "proxy-from-env";
 import {
   ACTOR_CONTEXT_HEADER,
@@ -172,6 +176,9 @@ export class CognigyApiClient {
           reqConfig,
           getProxyAxiosOptions(resolveRequestUrl(reqConfig)),
         );
+        // Redirects are followed inside the transport, below this interceptor,
+        // so each hop re-resolves its own route from its own destination.
+        reqConfig.beforeRedirect = applyProxyToRedirect;
         // Attribute this write to the plugin in Cognigy's audit events. Set
         // here rather than per call site so every platform request is covered,
         // including uploadFile's own headers object. Absent outside a tool
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -100,7 +100,7 @@ export function describeUnexpectedBody(
     message += `: the response did not match the expected Cognigy error format — ${snippet}`;
   }
 
-  const proxyUrl = safeGetProxyForUrl(resolveRequestUrl(error));
+  const proxyUrl = safeGetProxyForUrl(resolveRequestUrl(error.config));
   if (proxyUrl) {
     message +=
       ` This request went through the proxy ${redactProxyUrl(proxyUrl)};` +
@@ -113,13 +113,15 @@ export function describeUnexpectedBody(
 }
 
 /**
- * Best-effort absolute URL for the failed request. Request paths are relative
- * to the client's `baseURL`, and `getProxyForUrl` needs an absolute URL to
- * apply `NO_PROXY` — but neither field is guaranteed to be present or valid,
- * and an error path must never throw an error of its own.
+ * Best-effort absolute URL for a request. Request paths are usually relative to
+ * the client's `baseURL`, and `getProxyForUrl` needs an absolute URL to apply
+ * `NO_PROXY` — but neither field is guaranteed to be present or valid, and
+ * neither proxy selection nor the error path may throw an error of its own.
  */
-function resolveRequestUrl(error: AxiosError): string {
-  const { url, baseURL } = error.config ?? {};
+function resolveRequestUrl(
+  config: Pick<AxiosRequestConfig, "url" | "baseURL"> | undefined,
+): string {
+  const { url, baseURL } = config ?? {};
   try {
     if (url) return new URL(url, baseURL).toString();
   } catch {
@@ -152,14 +154,24 @@ export class CognigyApiClient {
         Accept: "application/json",
       },
       timeout: 30000,
-      // Every request to the platform shares one base URL, so the proxy (and
-      // any NO_PROXY exclusion) is resolved once here rather than per call.
-      ...getProxyAxiosOptions(config.baseUrl),
+      // Axios' own proxy handling never opens a CONNECT tunnel; the request
+      // interceptor below attaches real tunnelling agents instead.
+      proxy: false,
     });
 
     this.client.interceptors.request.use(
       (reqConfig) => {
         reqConfig.headers["X-API-Key"] = this.apiKey;
+        // Resolve the proxy against THIS request's target rather than the
+        // client's base URL. Not every request goes to the API host:
+        // downloadPackageArchive passes an absolute download link with
+        // `baseURL: undefined`, and that host has its own NO_PROXY standing —
+        // a client-wide decision would proxy it when it should not, or send it
+        // direct when the proxy is mandatory.
+        Object.assign(
+          reqConfig,
+          getProxyAxiosOptions(resolveRequestUrl(reqConfig)),
+        );
         // Attribute this write to the plugin in Cognigy's audit events. Set
         // here rather than per call site so every platform request is covered,
         // including uploadFile's own headers object. Absent outside a tool
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -7,6 +7,8 @@ import axios, {
 } from "axios";
 import FormData from "form-data";
 import { logger } from "../utils/logger.js";
+import { getProxyAxiosOptions, redactProxyUrl } from "../utils/proxy.js";
+import { getProxyForUrl } from "proxy-from-env";
 import {
   ACTOR_CONTEXT_HEADER,
   getActorContextHeader,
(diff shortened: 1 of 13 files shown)


src/utils/proxy.ts
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  try {
+    const agents = getAgents(proxyUrl);
+    logger.debug("Routing request through proxy", {
+      target: targetUrl,
+      proxy: redactProxyUrl(proxyUrl),
+    });
+    return { proxy: false, httpAgent: agents.http, httpsAgent: agents.https };
+  } catch (error: any) {
```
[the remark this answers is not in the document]
cu4nt0m: Fixed in e434b40.

src/api/client.ts
```diff
@@ -44,6 +46,88 @@ function isRetryable(error: AxiosError): boolean {
...
+    } catch {
+      snippet = String(data);
+    }
+  }
+  // Collapse whitespace: HTML error pages are mostly newlines and indentation,
+  // which would otherwise eat the snippet budget before the useful text.
+  snippet = snippet.replace(/\s+/g, " ").trim();
+  if (snippet.length > BODY_SNIPPET_LIMIT) {
+    snippet = `${snippet.slice(0, BODY_SNIPPET_LIMIT)}…`;
+  }
```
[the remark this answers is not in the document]
cu4nt0m: Fixed in e434b40.

src/api/client.ts
```diff
@@ -44,6 +46,88 @@ function isRetryable(error: AxiosError): boolean {
...
+
+  const parts = ["API request failed"];
+  const status = error.response?.status;
+  if (status) parts.push(`(HTTP ${status})`);
+
+  let message = parts.join(" ");
+  if (snippet) {
+    message += `: the response did not come from the Cognigy API — ${snippet}`;
+  }
```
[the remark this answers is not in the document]
cu4nt0m: Fixed in e434b40.

src/utils/proxy.ts
```diff
@@ -0,0 +1,190 @@
...
+ * every time.
+ */
+const agentCache = new Map<string, { http: Agent; https: Agent }>();
+
+function getAgents(proxyUrl: string): { http: Agent; https: Agent } {
+  let agents = agentCache.get(proxyUrl);
+  if (!agents) {
+    agents = {
+      http: new HttpProxyAgent(proxyUrl),
+      https: new HttpsProxyAgent(proxyUrl),
```
dshire: [P1] Bound proxy connection and CONNECT negotiation with a deadline

These agents do not return a socket to the request until proxy connection/CONNECT negotiation completes, while Axios's configured timeout starts only when the request receives its socket. A proxy that accepts the TCP connection but never answers CONNECT therefore leaves tool calls pending beyond the configured 30-second timeout. Reproduced with a local proxy that accepts and reads CONNECT without replying: an Axios request configured with timeout: 100 was still pending after 700 ms and required explicit cancellation. Please add a deadline covering proxy connection and tunnel negotiation, ensure pending sockets are cleaned up, and add a regression test with a stalled proxy.

src/api/client.ts
```diff
@@ -59,6 +152,9 @@ export class CognigyApiClient {
         Accept: "application/json",
       },
       timeout: 30000,
+      // Every request to the platform shares one base URL, so the proxy (and
+      // any NO_PROXY exclusion) is resolved once here rather than per call.
+      ...getProxyAxiosOptions(config.baseUrl),
```
dshire: [P2] Resolve proxy routing against each request's actual URL

The assumption that every request shares the API base URL does not hold for downloadPackageArchive: it calls this.apiClient.get(downloadLink, { baseURL: undefined, ... }) with an absolute download URL. If NO_PROXY excludes the API host but the download host requires a proxy, this instance retains proxy: false with no agents and sends the download directly; the inverse configuration routes an excluded download host through the proxy. Reproduced the first case with a local origin and proxy: getProxyAxiosOptions(downloadUrl) selected a proxy, but the client request reached the origin directly and the proxy received zero requests. Please select agents per request using its resolved target URL, preserving independent NO_PROXY/protocol handling for downloads.

src/utils/proxy.ts
```diff
@@ -0,0 +1,190 @@
...
+  if (rejection) {
+    // Warn and connect directly rather than tunnelling through something that
+    // cannot work: the agents accept almost any string and would silently
+    // build a tunnel to a nonsense host, turning a typo into a connection
+    // timeout with no clue as to why.
+    logger.warn(
+      `Ignoring proxy configuration (${rejection}); connecting directly instead`,
+      { proxy: redactProxyUrl(proxyUrl) },
+    );
+    return { proxy: false };
```
dshire: [P2] Reject unusable proxy settings instead of falling back to direct traffic

When a proxy is explicitly configured but malformed or unsupported, returning only { proxy: false } changes the request to a direct connection. Confirmed with a socks5:// setting: the warning was emitted and the request successfully reached the origin directly. This can send API credentials and request payloads outside the configured corporate proxy path, and on networks that prohibit direct traffic it replaces the useful configuration error with a downstream connection failure. Please surface an actionable configuration error and avoid issuing the request; apply the same behavior to the agent-construction catch below.

dshire (commented):
Review findings: three issues to address in the proxy implementation. Validated against e434b405 with local proxy reproductions. Typecheck, build, manifest checks, and tests passed after rerunning the sandbox-blocked test suite with loopback listening allowed.

src/utils/proxy.ts
cu4nt0m: Confirmed and fixed in eebbad9. Reproduced your case: axios' timeout is armed with `ClientRequest.setTimeout`, which starts only once the request has a socket, and the agent hands the socket over only after the tunnel is negotiated — so a silent proxy hung past the 30s timeout.

`withConnectDeadline` now races `connect()` against a deadline (default 30s, overridable with `COGNIGY_PROXY_CONNECT_TIMEOUT_MS`) and fails with `ETIMEDOUT`. On cleanup: the pending socket is aborted, not just orphaned. Both pinned agents call `net.connect(this.connectOpts)` synchronously before their first `await`, so an `AbortSignal` injected immediately around that call cannot interleave with another connect. Without that, the regression test's own server could not close — good signal that the fd really was being held.

Regression test in `src/__tests__/proxyRouting.test.ts`: a TCP server that reads CONNECT and never answers, with `timeout: 30000` on the request so only the deadline can end it. Fails in ~230ms.

src/api/client.ts
cu4nt0m: Confirmed and fixed in eebbad9. `downloadPackageArchive` (handlers.ts:1033) does pass an absolute link with `baseURL: undefined`, and resolving once from the client's base URL was wrong for it in both directions.

Proxy selection moved into the request interceptor and now resolves against each request's own resolved URL, so `NO_PROXY` and protocol handling apply per host. Regression test uses `127.0.0.1` for the API host and `localhost` for the download URL — same address, different hostnames, so `NO_PROXY=127.0.0.1` covers one and not the other. Asserts the API call reaches the origin with the proxy seeing zero requests, then that the download on the same client arrives at the proxy.

src/utils/proxy.ts
cu4nt0m: Agreed, and fixed in eebbad9. The argument that decided it: connecting directly sends the API key and payload outside the configured path, and on a network that forbids direct egress it trades an actionable configuration error for a confusing connection failure.

`getProxyAxiosOptions` now throws `ProxyConfigurationError` naming the proxy (credentials redacted) and pointing at `NO_PROXY` as the supported way to opt a host out. Same for the agent-construction catch, as you asked. Startup logs the problem once at error level so it is visible before the first tool call. Covered by unit tests plus an end-to-end one asserting a `socks5://` setting fails the request instead of reaching the origin.

dshire (commented):
Second pass: the three original reproductions are fixed. One remaining redirect-routing regression is detailed inline. Validation: 665 tests passed, 3 skipped; build and manifest checks passed.

src/api/client.ts
```diff
@@ -59,11 +154,24 @@ export class CognigyApiClient {
...
         reqConfig.headers["X-API-Key"] = this.apiKey;
+        // Resolve the proxy against THIS request's target rather than the
+        // client's base URL. Not every request goes to the API host:
+        // downloadPackageArchive passes an absolute download link with
+        // `baseURL: undefined`, and that host has its own NO_PROXY standing —
+        // a client-wide decision would proxy it when it should not, or send it
+        // direct when the proxy is mandatory.
+        Object.assign(
+          reqConfig,
+          getProxyAxiosOptions(resolveRequestUrl(reqConfig)),
```
dshire, line 177: [P2] Re-evaluate proxy routing for redirect destinations

Resolving agents in the request interceptor fixes separate requests to absolute download URLs, but Axios follows redirects inside its transport without running this interceptor again. A URL excluded by NO_PROXY that redirects to a host requiring the proxy therefore keeps the original direct route; conversely, a proxied request can keep using its proxy after redirecting to an excluded host. This can affect package downloads that redirect to the archive host. Confirmed with local servers and synthetic credentials: with NO_PROXY=127.0.0.1, an initial URL on 127.0.0.1 redirected to localhost; pre-PR Axios sent the redirected request through the configured proxy (direct hits: 0, proxy hits: 1), while this revision sent it directly (direct hits: 1, proxy hits: 0). Please resolve and replace the routing agents for each redirect destination as well, including clearing previously selected agents when the destination should be direct, and add redirect regression coverage.

cu4nt0m: Fixed in [8601b73](https://github.com/Cognigy/cognigy-plugin/pull/47/commits/8601b733d3858a38fb20e7d40f877b742ff792df)

dshire (approved):

